PR #1, opened 2024-01-28
Fischer Kolmogorov Solver Pagani Vignolo Hugot

LuigiPagani:
We've conducted the convergence analysis in 2D using a known solution, and the results appear to be solid. When applying the code to a realistic brain mesh with an identity matrix, the outcomes are also reasonably good. However, we've observed an issue when employing matrices such as Diag(0.1,10,0.1) or Diag(1,10,1): the solution, although largely reasonable and particularly coherent along the y-axis (the main direction of spread), exhibits slight negative values in some localized areas. We've noticed that these negative values become more pronounced as the ratio between the largest and smallest eigenvalues increases. Could this issue of negative values in certain regions be a cause for concern, or is it expected? 

Thanks for the availability.

Kind Regards

P.S.

The brain mesh is too big to be uploaded on GitHub.


LuigiPagani:
The brain mesh is too big to be uploaded on

michelebucelli:
> The brain mesh is too big to be uploaded on

As indicated in the repository README, please update the build and run instructions for your program with instructions on how to retrieve the mesh.

FK_solver/src/FK_solver.hpp
```diff
@@ -0,0 +1,282 @@
...
+    virtual double
+    value(const Point<dim> & /*p*/,
+          const unsigned int component_1 = 0,
+          const unsigned int component_2 = 1) const // override
+    {
+      if (component_1 == component_2)
+        return 1.0;
+      else
+        return 0.0;
+    }
```
michelebucelli: Here you are changing the interface of the class `dealii::Function` from which you are inheriting.

While C++ does admit this kind of thing, it is generally not a good idea from a design perspective, because it suggests that objects of type `FunctionD` can be used whenever an object of type `Function` is needed - which is not true.

To make an example, you could pass an object of type `FunctionD` to the function `dealii::VectorTools::interpolate` and everyhing would compile just fine. However, `interpolate` would try to call `FunctionD::value(const Point<dim>&, const unsigned int)` (notice the single `unsigned int` argument). This method is defined in the base class `Function`, and it is never overridden in the derived class. Hence, an user might think they were interpolating the tensor D, but in reality they are interpolating whatever the default implementation of `Function::value` contains. In medium-sized projects, this kind of thing will inevitably lead to hard-to-find bugs.

tl;dr: if you don't intend to use the interface of a certain class (as in this case), you should not inherit from that class.

FK_solver/src/main.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include <deal.II/base/convergence_table.h>
+
+
+// Main function.
+int
+main(int argc, char *argv[])
+{
+    Utilities::MPI::MPI_InitFinalize mpi_init(argc, argv);
+   
+  const unsigned int degree = 2;
```
michelebucelli: Indentation here is a bit messy. While it's not a huge problem for this file, since it only has a few lines, you should always aim for well-indented code, because it massively improves readability and eases debugging, especially in projects larger than this.

I suggest you look into automatic indenting tools such as `clang-format` (which incidentally can be called from within most modern text editors - including VSCode - so that you automatically indent your code whenever you save the source file).

FK_solver/src/FK_solver.cpp
```diff
@@ -0,0 +1,341 @@
...
+      fe_values.get_function_gradients(solution, solution_gradient_loc);
+      fe_values.get_function_values(solution_old, solution_old_loc);
+
+      for (unsigned int q = 0; q < n_q; ++q) {
+    // Evaluate coefficients on this quadrature node.
+    const double alpha_loc = alpha.value(fe_values.quadrature_point(q));
+
+    // Define the matrix to store D values and populate it.
+    FullMatrix<double> D_matrix(dim, dim);
+    D.matrix_value(fe_values.quadrature_point(q), D_matrix);
```
michelebucelli: Since you are already not using the `Function` interface, you can change this to an interface that uses the return value rather than input parameters:
```cpp
const FullMatrix<double> D_matrix = D.matrix_value(fe_values.quadrature_point(q));
```
This is generally more modern and readable, and thanks to return value optimization it should have zero overhead.

If instead you want to be sure not to allocate (and then de-allocate) a new `FullMatrix` for every quadrature node, you should keep the interface as it is, but instead define `D_matrix` outside the loop on elements, and reuse it for every iteration.

FK_solver/src/FK_solver.hpp
```diff
@@ -0,0 +1,282 @@
...
+  };
+
+  // Function of the matrix D
+  class FunctionD : public Function<dim>
+  {
+  public:
+    virtual void
+    matrix_value(const Point<dim> & /*p*/,
+                 FullMatrix<double> &values) const // override
```
michelebucelli: Mathematically speaking, $D$ is a **tensor**, which is represented by a matrix for a given coordinate system. This is the same (admittedly subtle) difference that exists between an n-dimensional vector (in the physical sense) and its representation through the n-by-1 matrix of its components with respect to a given coordinate system.

`deal.II` uses the class template `Tensor` to represent such objects: `Tensor<1, dim>` for `dim`-dimensional physical vectors (e.g. gradients, or velocities) and `Tensor<2, dim>` for `dim`-dimensional tensors of rank 2.

I suggest you use `Tensor<2,dim>` in place of `FullMatrix<double>`. On top of many semantical benefits, this will allow you to compute $D\nabla\varphi_j$ with the multiplication operator `*`, without the need to loop over components.

FK_solver/src/FK_solver.cpp
```diff
@@ -0,0 +1,341 @@
...
+                                fe_values.JxW(q);
+
+            cell_matrix(i, j) += 2.0 * alpha_loc *
+                                 solution_loc[q] * 
+                                 fe_values.shape_value(j, q) * 
+                                 fe_values.shape_value(i, q) * 
+                                 fe_values.JxW(q);
+
+            Tensor<1, dim> grad_phi_i = fe_values.shape_grad(i, q);
+            Tensor<1, dim> grad_phi_j = fe_values.shape_grad(j, q);
```
michelebucelli: This is actually fairly inefficient: you are allocating and deallocating two tensors for every element, for every quadrature node, for every pair of local DoFs $(i, j)$. The method `FEValues::shape_grad` returns its result by `const` reference precisely to avoid this unnecessary allocation and deallocation. If you want to alias the result for the sake of readability, you can just use
```cpp
const Tensor<1, dim> &grad_phi_i = fe_values.shape_grad(i, q);
const Tensor<1, dim> &grad_phi_j = fe_values.shape_grad(j, q);
```
(emphasis on the variables being const **references** - no copy is made here, you're just giving a new name to the variables returned by `shape_grad`).

FK_solver/src/FK_solver.cpp
```diff
@@ -0,0 +1,341 @@
...
+            //compute the matrix vectol product
+            for (unsigned int d = 0; d < dim; ++d) {
+              //D_grad_phi_j[d] = 0.0;
+              for (unsigned int e = 0; e < dim; ++e) {
+                D_grad_phi_j[d] += D_matrix(d, e) * grad_phi_j[e];
+              }
+            }
+
+            cell_matrix(i, j) += scalar_product(D_grad_phi_j, grad_phi_i) *
+                                 fe_values.JxW(q);
```
michelebucelli: See my other comment on using `Tensor` to define the matrix $D$. All these lines would become
```cpp
cell_matrix(i, j) += scalar_product(D * grad_phi_j, grad_phi_i) * fe_values.JxW(q);
```
The only possible downside of this approach is that it requires the allocation of a temporary `Tensor<1, dim>` to store the result of `D * grad_phi_j`. However, this is usually an acceptable price to pay for the simplicity of the implementation (which entails readability and maintainability in the long run).

If (and only if) you should find out that this line becomes a major performance bottleneck, the following is a possible optimization:
```cpp
for (unsigned int c = 0; c < dim; ++c) {
  for (unsigned int d = 0; d < dim; ++d) {
    cell_matrix(i, j) += grad_phi_j[c] * D_matrix[c][d] * grad_phi_i[d];
  }
}
```
(where the syntax assumes that `D_matrix` is now a `Tensor<2, dim>`).

michelebucelli (commented):
Thanks for submitting! I have left a few suggestions above. Feel free to ask any questions or clarifications.

LuigiPagani:
> > The brain mesh is too big to be uploaded on
> 
> As indicated in the repository README, please update the build and run instructions for your program with instructions on how to retrieve the mesh.

Dear Professor, thanks a lot for the suggestions. We should have fixed these things in the last commit.

LuigiPagani:
> Thanks for submitting! I have left a few suggestions above. Feel free to ask any questions or clarifications.

Dear Professor, thanks a lot for the suggestions. We have tried to implement them in the last commit.

Kind Regards,

Luigi Pagani, Samuele Vignolo, Simone Hugot

PR #2, opened 2024-01-31
Project 3 - Allahakbari Miotti Pesce

FrancescoPesce:
Dear Prof. Bucelli, as requested, we are opening a pull request to the repository, but might continue updating the code until the extended deadline.


README.md
```diff
@@ -17,7 +10,84 @@ $ cd build
...
+- **`-u, --inlet-velocity` <u>**: specifies the reference velocity of the inflow. This parameter is only relevant for the 2D and 3D flow past a cylinder problems.
+- **`-v, --varying-inlet`**: this is a boolean flag. If provided, the inflow velocity will change in time. This parameter is only relevant for the 2D and 3D flow past a cylinder problems.
+- **`-i, --no-inner-solver`**: use a single sweep of the preconditioner instead of an inner solver for the aSIMPLE preconditioner. This parameter is only relevant if aSIMPLE is selected.
+- **`-l, --ilu-preconditioner`**: use an ILU preconditioner instead of AMG for the inner solvers.
+- **`-a, --alpha` <a>**: value of alpha for the SIMPLE or aSIMPLE preconditioners, it must be greater than 0 and at most 1, which is the default value. This parameter is only relevant if SIMPLE or aSIMPLE are selected.
+- **`-x --tol` <tol>**: relative tolerance for the main solver, the default value is `1e-7`.
+- **`-y --tol-inner` <tol-inner>**: relative tolerance for the inner solvers, the default value is `1e-5`. Some preconditioners might need this value to be below a certain amount in some problems.
```
michelebucelli, line 47: Having these parameters set from the command line is neat. However, as the number of parameters grows, it can become slightly cumbersome. I suggest you look into the `dealii::ParameterHandler` class for a more structured and scalable way of doing this.

src/main.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+      case '?':
+        pcout << err_msg << std::endl;
+        return 1;
+
+      default:
+        pcout << err_msg << std::endl;
+        return 1;
+    }
+  }
```
michelebucelli, line 168: You might also consider `clipp` for a more modern way of managing command line options.

src/main.cpp
```diff
@@ -0,0 +1,333 @@
...
+        const double T = 4.0 * deltat;
+        constexpr double nu = 0.01;
+
+        std::vector<double> h_values;
+        std::vector<double> errors_L2;
+        std::vector<double> errors_H1;
+
+        for (unsigned int i = 0; i < num_files; i++) {
+          // Extract the factor from the file name
+          size_t pos = mesh_file_name.find_last_of('-');
```
michelebucelli, line 243: You should always declare a variable as `const` if it is not meant to change:
```suggestion
          const size_t pos = mesh_file_name.find_last_of('-');
```

src/main.cpp
```diff
@@ -0,0 +1,333 @@
...
+            std::cerr << "Invalid file format. Exiting." << std::endl;
+            return 1;
+          }
+
+          std::string factorString = mesh_file_name.substr(pos + 1);
+
+          double mesh_factor = std::stod(factorString);
+
+          // Halve the factor
+          mesh_factor /= pow(2, i);
```
michelebucelli, line 254: In the same vein as before:
```suggestion
          const std::string factorString = mesh_file_name.substr(pos + 1);
          const double mesh_factor = std::stod(factorString) / std::pow(2, i);
```

include/SolverOptions.hpp
```diff
@@ -0,0 +1,63 @@
+#ifndef SOLVER_OPTIONS_HPP
+#define SOLVER_OPTIONS_HPP
+
+// An enumeration of possible preconditioners for the problem.
+enum preconditioner_id { BLOCK_DIAGONAL, SIMPLE, ASIMPLE, YOSHIDA };
```
michelebucelli, line 5: In modern C++, it is generally better to use an `enum class` rather than a simple `enum`:
```suggestion
enum class preconditioner_id {
  BLOCK_DIAGONAL,
  SIMPLE,
  ASIMPLE,
  YOSHIDA
};
```
The `enum class` offers a good deal of semantic advantages with respect to the `enum` (e.g. it is not implicitly casted to the underlying integer type).

(on a minor note, the preconditioner is actually called "Yosida").

src/Assemble.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+    // Declare tensors to store the old solution value at a set quadrature
+    // point and part of the nonlinear term.
+    Tensor<1, dim> local_old_solution_value;
+    Tensor<1, dim> nonlinear_term;
+
+    for (const auto &cell : dof_handler.active_cell_iterators()) {
+      if (!cell->is_locally_owned()) continue;
+
+      fe_values.reinit(cell);
```
michelebucelli, line 157: The loop over elements and the reinitialization of the `FEValues` instance is generally quite costly. Here you're looping twice for no particular reason (once for the matrix and once for the right-hand side), which is less efficient than doing a single loop.

src/Assemble.cpp
```diff
@@ -0,0 +1,296 @@
...
+  {
+    std::map<types::global_dof_index, double> boundary_values;
+
+    // Since Dirichlet (BC) are not applied to the pressure, create a mask.
+    ComponentMask mask;
+    if constexpr (dim == 2) {
+      mask = ComponentMask({true, true, false});
+    } else {
+      mask = ComponentMask({true, true, true, false});
+    }
```
michelebucelli, line 275: This can be done more elegantly as follows:
```suggestion
    ComponentMask mask(dim + 1, true);
    mask.set(dim, false);
```

src/Assemble.cpp
```diff
@@ -0,0 +1,296 @@
...
+        const_dirichlet_boundary_functions;
+
+    // Set the correct time for the Dirichlet functions and create a constant
+    // copy of the map (needed because interpolate_boundary_values requires a
+    // map with constant values).
+    for (auto iter = dirichlet_boundary_functions.begin();
+         iter != dirichlet_boundary_functions.end(); ++iter) {
+      iter->second->set_time(time_step * deltat);
+      const_dirichlet_boundary_functions[iter->first] =
+          const_cast<const Function<dim> *>(iter->second);
```
michelebucelli, line 287: I'm pretty sure that you don't need this `const_cast` - any non-`const` pointer can be converted to its `const` counterpart.

src/Solve.cpp
```diff
@@ -0,0 +1,132 @@
...
+void NavierStokes<dim>::apply_initial_conditions() {
+  pcout << "  Applying the initial conditions" << std::endl;
+
+  // Since only velocity has an initial condition, create a mask.
+  ComponentMask mask;
+  if constexpr (dim == 2) {
+    mask = ComponentMask({true, true, false});
+  } else {
+    mask = ComponentMask({true, true, true, false});
+  }
```
michelebucelli, line 22: Same comment as elsewhere.

src/Solve.cpp
```diff
@@ -0,0 +1,132 @@
...
+          std::make_shared<PreconditionYoshida>();
+      actual_precondition->initialize(
+          system_matrix.block(0, 0), system_matrix.block(1, 0),
+          system_matrix.block(0, 1), velocity_mass.block(0, 0), solution_owned,
+          solver_options.maxiter_inner, solver_options.tol_inner,
+          solver_options.use_ilu);
+      precondition = actual_precondition;
+      break;
+    }
+  }
```
michelebucelli, line 83: If I understand correctly, you use two pointers (one to the base class, one to the derived class) because the base class does not have an `initialize` method.

However, seeing as you call `initialize` right after constructing the object, you could just merge the `initialize` method with the constructor of each derived class (the constructors of classes derived from the same base can have different signatures). This way, for every `case` statement you would only have to write
```cpp
case BLOCK_DIAGONAL:
  preconditioner = std::make_shared<PreconditionBlockDiagonal>( /* ...args... */);
  break;
```

A couple of side notes:
- you should not use a `shared_ptr` when a `unique_ptr` would do (as is the case here);
- your `switch` statements should always have a `default` case, and if you have no way of handling the default case you should throw an exception.

src/Output.cpp
```diff
@@ -0,0 +1,42 @@
...
+          dim, DataComponentInterpretation::component_is_part_of_vector);
+  data_component_interpretation.push_back(
+      DataComponentInterpretation::component_is_scalar);
+
+  std::vector<std::string> names;
+  if constexpr (dim == 2) {
+    names = {"velocity", "velocity", "pressure"};
+  } else {
+    names = {"velocity", "velocity", "velocity", "pressure"};
+  }
```
michelebucelli, line 24: This can also be done more elegantly:
```cpp
std::vector<std::string> names(dim, "velocity");
names.push_back("pressure");
```

include/Cylinder.hpp
```diff
@@ -0,0 +1,268 @@
...
+template <unsigned int dim>
+class Cylinder : public NavierStokes<dim> {
+ public:
+  // Virtual destructor.
+  virtual ~Cylinder() = default;
+
+  // Update lift and drag forces. In particular, this function computes the
+  // lift and drag coefficients and prints them to the console. These
+  // coefficients will be added to a .csv file and plotted in a Python script.
+  void update_lift_drag();
```
michelebucelli, line 28: It is generally advisable to separate the computations from the output. In this case, for instance, at some point in the future you may want to use the drag and lift coefficients e.g. in an optimal control loop.

Therefore, from a design perspective, it would probably be better to have this method return its result (e.g. as an `std::pair<double, double>`) instead of writing it to the standard output and/or to a CSV file. The output can be managed separately (e.g. in the output method itself).

src/Cylinder.cpp
```diff
@@ -0,0 +1,181 @@
...
+                        const SolverOptions &solver_options_,
+                        const bool &compute_lift_drag_)
+    : NavierStokes<dim>(mesh_file_name_, degree_velocity_, degree_pressure_, T_,
+                        deltat_, solver_options_),
+      U_m(U_m_),
+      compute_lift_drag(compute_lift_drag_),
+      zero_function(dim + 1) {
+  NavierStokes<dim>::ro = 1.0;   //[kg/m^3].
+  NavierStokes<dim>::nu = 1e-3;  //[m^2/s].
+  NavierStokes<dim>::initial_conditions =
```
michelebucelli, line 22: You don't really need to specify the scope `NavierStokes<dim>::` in classes derived from `NavierStokes` (such as this one). It is only needed if there's a name clash (e.g. you have a member called `ro` in both `NavierStokes<dim>` and `Cylinder<dim>`), but this is not the case here.

src/LiftDrag.cpp
```diff
@@ -0,0 +1,439 @@
...
+          const Tensor<1, dim> &neg_normal = fe_face_values.normal_vector(q);
+
+          // Calculate the stress tensor.
+          stress_tensor = velocity_gradients[q];
+          for (unsigned int i = 0; i < dim; i++) {
+            for (unsigned int j = 0; j < dim; j++) {
+              stress_tensor[i][j] += velocity_gradients[q][j][i];
+            }
+          }
+          stress_tensor *= NavierStokes<dim>::nu;
```
michelebucelli, line 128: The following would probably be more readable:
```suggestion
          stress_tensor = nu * (velocity_gradients[q] + transpose(velocity_gradients[q]);
```
Admittedly, it may be slightly less efficient, due to the need of allocating and de-allocating the temporary `Tensor<2, dim>` needed to store the result of the sum. If such an inefficienty became a bottleneck, the following would still improve readability over the currently implemented version:
```suggestion
          const auto & grad_u = velocity_gradients[q];
          for (unsigned int i = 0; i < dim; i++) {
            for (unsigned int j = 0; j < dim; j++) {
              stress_tensor[i][j] = nu * (grad_u[i][j] + grad_u[j][i]);
            }
          }
```
Notice how the line where the tensor is actually built closely resembles its mathematical definition.

michelebucelli (commented):
Thanks for submitting your project! I have left a few comments and suggestions above.

On a general note, I find it is generally preferrable to keep the implementation of a class in a single file, unless it is very long (maybe a few thousand of lines) and thus hard to navigate. For instance, I would have expected to have the implementation of `NavierStokes` into a single source file, e.g. `NavierStokes.cpp`, instead of having it split over multiple file.

While separating the implementation over multiple files may be helpful as long as you know where the different pieces are, keeping track of this may become cumbersome for larger projects, and is definitely not easy for someone who reads your code for the first time.

PR #4, opened 2024-02-01
Project of the group: Liborio Roman, Kamil Hanna and Filippo Mininno

FilippoIspanico:
# Preconditioning heterogeneous diffusion problems
By Liborio Roman, Kamil Hanna and Filippo Mininno


src/Poisson3D.cpp
```diff
@@ -0,0 +1,460 @@
...
+                return;
+
+            default:
+                std::cout<<"default"<<std::endl;
+                break;
+        }
+
+    }
+
+   mesh_file_name = "../mesh/" + mesh_name_no_path;
```
michelebucelli, line 113: I quite appreciate the inclusion of a command line interface. For future projects, you might consider looking into more friendly tools for this purpose, such as [clipp](https://github.com/muellan/clipp).

src/Poisson3D.hpp
```diff
@@ -0,0 +1,303 @@
...
+      std::vector<Sphere> spheres;
+      double p_value;
+
+      bool is_point_inside_sphere(const Point<dim>& point, const Sphere& sphere) const
+      {
+          double distance_squared = 0.0;
+          for (unsigned int i = 0; i < dim; ++i)
+          {
+              distance_squared += std::pow(point[i] - sphere.center[i], 2);
+          }
```
michelebucelli, line 100: This can be done more expressively and conveniently using `dealii::Point<dim>::distance_square`.

src/Poisson3D.hpp
```diff
@@ -0,0 +1,303 @@
...
+          return distance_squared <= std::pow(sphere.radius, 2);
```
michelebucelli, line 101: You should **never** use  `std::pow` for integer powers, because it is hugely inefficient with respect to `sphere.radius * sphere.radius`. `std::pow` is meant to compute non-integer powers.

src/Poisson3D.hpp
```diff
@@ -0,0 +1,303 @@
...
+          }
+          return 1.0;
+      }
+
+  private:
+      std::vector<Sphere> spheres;
+      double p_value;
+
+      bool is_point_inside_sphere(const Point<dim>& point, const Sphere& sphere) const
+      {
```
michelebucelli, line 95: Wouldn't it make more sense for this to be a method of `Sphere`? You would then be able to simply call `sphere.is_inside(point)`.

src/Poisson3D.hpp
```diff
@@ -0,0 +1,303 @@
...
+          {
+              if (is_point_inside_sphere(p, sphere))
+                  return std::pow(10.0, p_value);
+          }
+          return 1.0;
+      }
+
+  private:
+      std::vector<Sphere> spheres;
+      double p_value;
```
michelebucelli, line 92: If data members are not supposed to change over time, as is the case here, they should be declared as const:
```suggestion
      const std::vector<Sphere> spheres;
      const double p_value;
```

src/Poisson3D.hpp
```diff
@@ -0,0 +1,303 @@
...
+            Point<dim> center = {dis_center(gen), dis_center(gen), dis_center(gen)};
+            double radius = dis_radius(gen);
+            spheres.push_back({center, radius});
+        }
+
+
+
+        // Initialize the diffusion coefficient
+        diffusion_coefficient = DiffusionCoefficient<dim>(spheres, p_value);
+  };
```
michelebucelli, line 194: Were the spheres placed in different positions for the different tests? If so, this migth significantly affect the results.

While in principle having randomly-varying coefficients is not a problem, it is generally unwise to compare e.g. the compute times or the number of linear solver iterations between tests with different setups.

You should consider making sure that the random number generator is always initialized with the same seed.

src/Poisson3D.hpp
```diff
@@ -0,0 +1,303 @@
...
+      dealii::Vector<double> x(matrix.n()), b(matrix.m());
+      b = 1; // Or any other non-zero vector
+
+      try {
+          solver.solve(matrix, x, b, dealii::PreconditionIdentity());
+          return false;
+      } catch (std::exception &e) {
+          return true; // The matrix could be singular or poorly conditioned
+      }
+  }
```
michelebucelli, line 173: For methods longer than a few lines, it is generally preferrable to write their definition in the *.cpp file, and only their declaration in the *.hpp file.

src/Poisson3D.hpp
```diff
@@ -0,0 +1,303 @@
...
+};
+
+  void
+  manage_flags(int argc, char ** argv);
+
+  void
+  write_csv(const long int elapsed_time, int iterations) const;
+
+  static std::string
+  extractFileName(const std::string& filePath);
```
michelebucelli, line 228: These methods have no documentation.

Writing the code documentation is generally a good idea, so that you are explicitly aware of what you intended your code to do, and that you communicate this to other people reading the code.

src/Poisson3D.cpp
```diff
@@ -0,0 +1,460 @@
...
+             << std::endl;
+    csv_file.close();
+
+    std::cout<< "CSV write successful."<<std::endl;
+}
+
+std::string
+Poisson3D::extractFileName(const std::string& filePath) {
+
+    size_t lastSlash = filePath.find_last_of("/\\");
```
michelebucelli, line 31:
```suggestion
    const size_t lastSlash = filePath.find_last_of("/\\");
```
Always declare as const the variables whose value is not meant to change.

src/Poisson3D.cpp
```diff
@@ -0,0 +1,460 @@
...
+
+
+    std::string fileName = filePath.substr(lastSlash + 1);
+
+
+    size_t lastDot = fileName.find_last_of(".");
+
+
+    return fileName.substr(0, lastDot);
+}
```
michelebucelli, line 41: This can almost certainly be done using the standard library header `filesystem`.

src/Poisson3D.cpp
```diff
@@ -0,0 +1,460 @@
...
+  //Set to get the boundaryIds
+  std::set<types::boundary_id> boundary_ids;
+
+  for (const auto &cell : dof_handler.active_cell_iterators())
+    {
+      for (const auto &face : cell->face_iterators()) {
+        if (face->at_boundary()) {
+            boundary_ids.insert(face->boundary_id());
+        }
+    }
```
michelebucelli, line 273: This piece of code is indented poorly, and in the worst possible way.

The closing bracket on line 273 is vertically aligned to the opening bracket of line 268, thus suggesting on a first look that the for loop starting on line 267 ends here. However, the bracket of line 273 actually closes the for loop of line 269.

This sort of thing makes the code incredibly hard to read. You should always correctly indent the code. You should consider looking into automatic indentation tools such as clang-format. You can also configure your editor so that automatic indentation is triggered upon saving the source code file.

src/Poisson3D.cpp
```diff
@@ -0,0 +1,460 @@
...
+          for (unsigned int i = 0; i < dofs_per_cell; ++i)
+            {
+              for (unsigned int j = 0; j < dofs_per_cell; ++j)
+                {
+                  // FEValues::shape_grad(i, q) returns the gradient of the i-th
+                  // basis function at the q-th quadrature node, already mapped
+                  // on the physical element: we don't have to deal with the
+                  // mapping, it's all hidden inside FEValues.
+                  cell_matrix(i, j) += diffusion_coefficient.value(
+                                         fe_values.quadrature_point(q)) // mu(x)
```
michelebucelli, line 301: This may be slightly inefficient, since the evaluation of the diffusion coefficient is not trivial.

Indeed, for every quadrature node, you are evaluating the diffusion coefficient `dofs_per_cell*dofs_per_cell`, instead of once. You should consider moving this evaluation to the outer loop (the one on `q`).

src/Poisson3D.cpp
```diff
@@ -0,0 +1,460 @@
...
+Poisson3D::solve()
+{
+  std::cout << "===============================================" << std::endl;
+
+  // Here we specify the maximum number of iterations of the iterative solver,
+  // and its tolerance.
+  SolverControl solver_control(10000, 1e-6 * system_rhs.l2_norm());
+  for (unsigned int i = 0; i < solution.size(); ++i) {
+      solution[i] = 0.0;
+  }
```
michelebucelli, line 362: You can just write `solution = 0.0;`, since the `Vector` class has an assignment operator for this purpose.

michelebucelli (commented):
Thanks for submitting your project! Please find some comments and suggestions above.

PR #5, opened 2024-02-01
PROJECT 3 Cenzato Pisante Procaccio

mattecenz:
Here is the pull request of the project. In the README there is a small description on how to run the tests and where to find all the commit history if needed. 

Hopefully no more commits to the code will be performed until tomorrow, maybe the only thing that we are adding is the full report, which will be also sent via e-mail.


mattecenz:
I noticed just one small detail. When launching the 3D problem it will default to the 3D-unsteady problem, which as explained in the report does not converge. To modify it one has to manually change the inlet velocity to a more suitable value.

src/main2D.cpp
```diff
@@ -0,0 +1,56 @@
+#include "NavierStokes2D.hpp"
+
+// Main function.
+int main(int argc, char *argv[])
+{
+  Utilities::MPI::MPI_InitFinalize mpi_init(argc, argv);
+
+  int rank;
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
```
michelebucelli, line 10: This is definitely correct, albeit a bit old fashioned. `deal.II` offers functions with a more modern interface to this purpose within the namespace `Utilities::MPI` (in this case, `Utilities::MPI::this_mpi_process`).

This allows you to write statements like
```cpp
const int rank = Utilities::MPI::this_mpi_process(MPI_COMM_WORLD);
```
which is generally safer than what you have in terms of constness and correct initialization of the variable `rank`.

src/main2D.cpp
```diff
@@ -0,0 +1,56 @@
...
+  const std::string mesh_file_name = argc > 1 ? argv[1] : "../mesh/cilinder_2D_fine.msh";
+  const unsigned int degree_velocity = 2;
+  const unsigned int degree_pressure = 1;
+
+  const double T = 8;
+  const double deltat = 0.005;
+
+  dealii::Timer timer;
+  // Start the timer
+  timer.restart();
```
michelebucelli, line 20: This timer may not work appropriately in parallel.

Suppose that the load is not well distributed among processes. Each process will measure the time it takes to perform its part of the solution. Then, rank 0 will report the time it took it to perform its part of the solution. If, due to unbalancing issues, rank 0 has significantly less work than the others, this timer will underestimate the actual time needed to compute the solution.

In practice, things are not that bad because within both `setup` and `solve` you probably have (implicit) MPI barriers that force the processes to sync up. However, it is not really a good design to base your measurements on the existence of these barriers defined elsewhere.

A better approach is to pass an MPI communicator and set `sync_lap_times` to true in the constructor of `Timer`. An even better approach is to use the `dealii::TimerOutput` class, also making sure to initialize it correctly for parallel programs.

src/main2D.cpp
```diff
@@ -0,0 +1,56 @@
...
+    {
+      std::cerr << "Error opening output file" << std::endl;
+      return -1;
+    }
+    outputFile << "Iteration, Drag, Lift, Coeff Drag, CoeffLift, time prec, time solve" << std::endl;
+
+    for (size_t ite = 0; ite < problem.vec_drag.size(); ite++)
+    {
+      outputFile << ite * deltat << ", " << problem.vec_drag[ite] << ", " << problem.vec_lift_coeff[ite] << ", " 
+                << problem.vec_drag_coeff[ite] << ", " << problem.vec_lift_coeff[ite] << ", "
```
michelebucelli: It seems to me that you are exporting the lift coefficient twice, instead of the lift force and lift coefficient. Maybe there's a typo here?

src/main3D.cpp
```diff
@@ -0,0 +1,23 @@
...
+main(int argc, char *argv[])
+{
+  Utilities::MPI::MPI_InitFinalize mpi_init(argc, argv);
+
+  const std::string  mesh_file_name  = argc>1?argv[1]:"../mesh/cilinder_3D_fine.msh";
+  const unsigned int degree_velocity = 2;
+  const unsigned int degree_pressure = 1;
+
+	const double T      = 8.0;
+	const double deltat = 0.05;
```
michelebucelli: These lines are not indented properly.

This is clearly a minor issue here, because this is a small file with a very simple function, and because this is a very small project. However, issues like this can become very dangerous in larger projects, as they may hide bugs and overall hinder the readability, and thus the maintainability.

I suggest you look into automatic indentation software such as `clang-format`, which has very nice plugins for most text editors, allowing the automatic indentation of the source code whenever it is saved.

src/NavierStokes2D.hpp
```diff
@@ -0,0 +1,633 @@
...
+
+    double getMeanVelocity() const
+    {
+
+      return 8 * u_m * H / 2 * H / 2 / (3.0 * H * H);
+    }
+
+  protected:
+    double H = 0.41;
+    double u_m = 0.3; // 0.3;
```
michelebucelli: I assume these two values are not expected to change over time. You should make them const:
```suggestion
    const double H = 0.41;
    const double u_m = 0.3; // 0.3;
```
Moreover, these are actually compile-time constants. It is probably better to declare them as such:
```suggestion
    static inline constexpr double H = 0.41;
    static inline constexpr  u_m = 0.3; // 0.3;
```
This will help the compiler in optimizing these variables out whenever it can.

src/NavierStokes2D.hpp
```diff
@@ -0,0 +1,633 @@
...
+    {
+      F = &F_;
+      B = &B_;
+      B_T = &B_t;
+
+      diag_D_inv.reinit(sol_owned.block(0));
+
+      for (unsigned int i : diag_D_inv.locally_owned_elements())
+      {
+        double temp = F->diag_element(i);
```
michelebucelli: Whenever you know that a variable is not supposed to change, you should declare it `const`:
```suggestion
        const double temp = F->diag_element(i);
```
However, I think that computing the inverse in one line would probably work just as fine here:
```cpp
diag_D_inv[i] = 1.0 / F->diag_element(i);
```

src/NavierStokes2D.hpp
```diff
@@ -0,0 +1,633 @@
...
+  void
+  solve();
+
+  std::vector<double> vec_drag;
+  std::vector<double> vec_lift;
+  std::vector<double> vec_drag_coeff;
+  std::vector<double> vec_lift_coeff;
+
+  std::vector<double> time_prec;
+  std::vector<double> time_solve;
```
michelebucelli: Exposing members from a class in the public scope is generally regarded as a bad design decision, unless those members have absolutely no constraints on their value.

To make an example where this would make sense: suppose you are writing a class for complex numbers. Then, the real and imaginary parts would probably be two data members, and it would make sense for them to take any value, without any constraint.

Here, instead, the `vec_*` coefficients are all meant to store exactly as many elements as there are timesteps in your simulation (which additionally implies the constraint that they should all have the same length). Furthermore, their content is not arbitrary, but should be precisely the lift (or drag) force (or coefficient).

Since all these are public members, any user of your class can do whatever they want with them, including changing their lengths and changing or erasing their contents, thus rendering the state of this class invalid. The whole purpose of having a class is that you want to make sure that your objects always remain in a valid state.

tl;dr: these should be made `protected` or `private`, and exposed through a `const` getter method.

src/NavierStokes2D.cpp
```diff
@@ -0,0 +1,565 @@
...
+  // PreconditionBlockIdentity preconditioner;
+
+  // PreconditionSIMPLE preconditioner;
+
+  PreconditionaSIMPLE preconditioner;
+
+  pcout << " Assemblying the preconditioner... " << std::endl;
+
+  dealii::Timer timerprec;
+  timerprec.restart();
```
michelebucelli: Same observation as before on the use of this class in parallel.

src/NavierStokes2D.cpp
```diff
@@ -0,0 +1,565 @@
...
+          fe_face_values.reinit(cell, f);
+
+          for (unsigned int q = 0; q < n_q_face; ++q)
+          {
+            // Get the values
+            const double nx = fe_face_values.normal_vector(q)[0];
+            const double ny = fe_face_values.normal_vector(q)[1];
+
+            // Construct the tensor
+            Tensor<1, dim> tangent;
```
michelebucelli: You might consider allocating this object only once outside of the loop.

src/NavierStokes2D.cpp
```diff
@@ -0,0 +1,565 @@
...
+        }
+      }
+    }
+  }
+  drag = Utilities::MPI::sum(local_drag, MPI_COMM_WORLD);
+  lift = Utilities::MPI::sum(local_lift, MPI_COMM_WORLD);
+  pcout << "Drag :\t " << drag << " Lift :\t " << lift << std::endl;
+  // The mean velocity is defined as 2U(0,H/2,t)/3
+  // This is in the case 2D-2 unsteady
+  double mean_v = inlet_velocity.getMeanVelocity();
```
michelebucelli:
```suggestion
  const double mean_v = inlet_velocity.getMeanVelocity();
```

michelebucelli (commented):
Thanks for delivering your project. I have left a few comments above. A few general observations follow.

1. Have you considered making the solver a class template with respect to the physical dimension `dim`? The whole purpose of dimension-independent code (i.e. of having code where the dimension can be changed by simply changing the value of a variable) is that you don't need to rewrite the same code multiple times, once for 2D and once for 3D.
2. There are some parts of your code that are quite lacking in comments and documentation. For instance, the meaning of the data members `vec_*` is not documented and must be inferred from the context. This is not a good idea - comments and documentation are important first and foremost for the developer themselves, since they allow them to lay down the specification of what they're trying to do.
3. From a design perspective, there are quite a few hardcoded constants, and generally the Navier-Stokes solver is quite specific for one particular problem (e.g. it imposes boundary conditions on some specific boundaries, etc.). While this is not a problem per se, at least for the purpose of this project, it might be a good exercise to try and separate the two concerns of solving the NS equations themselves and of imposing boundary conditions. For instance, you could consider having some methods that pass boundary conditions to your class from the outside.

PR #8, opened 2024-06-04
Matrix Free Solver by Greco Guerrini Marzo

LukeTheWalker:
Here the PR of our project of NMPDE for next exam session


src/DTR.cpp
```diff
@@ -0,0 +1,424 @@
...
+    system_matrix.reinit(sparsity);
+
+    // Finally, we initialize the right-hand side and solution vectors.
+    pcout << "  Initializing the system right-hand side" << std::endl;
+    system_rhs.reinit(locally_owned_dofs, MPI_COMM_WORLD);
+    pcout << "  Initializing the solution vector" << std::endl;
+    solution.reinit(locally_owned_dofs, MPI_COMM_WORLD);
+  }
+
+  setup_time += time.wall_time();
```
michelebucelli, line 112: Have you considered using the class `dealii::TimerOutput::Scope` to evaluate the wall time of different code portions? I find it allows for a considerably nicer profiling experience than manually keeping track of wall times.

src/DTR.cpp
```diff
@@ -0,0 +1,424 @@
...
+    {
+      // Here we assemble the local contribution for current cell and
+      // current quadrature point, filling the local matrix and vector.
+      Vector<double> b_loc(dim);
+
+      transport_coefficient.vector_value(fe_values.quadrature_point(q), b_loc);
+
+      Tensor<1, dim> b_loc_tensor;
+      for (unsigned int i = 0; i < dim; ++i)
+        b_loc_tensor[i] = b_loc[i];
```
michelebucelli, line 192: I believe you have a `tensor_value` method that would save you this conversion.

src/DTR.cpp
```diff
@@ -0,0 +1,424 @@
...
+    // condition is u_i = b, the map will contain the pair (i, b).
+    std::map<types::global_dof_index, double> boundary_values;
+
+    // Then, we build a map that, for each boundary tag, stores the
+    // corresponding boundary function.
+    std::map<types::boundary_id, const Function<dim> *> boundary_functions;
+
+    if (problem_data::bcs[0] != 'D' || problem_data::bcs[2] != 'D') { std::cerr << "Dirichlet boundary conditions are not set correctly" << std::endl; exit(1); }
+    boundary_functions[0] = &dirichletBC1;
+    boundary_functions[2] = &dirichletBC2;
```
michelebucelli, line 294: This looks a bit confusing.

On the one hand you have a variable `problem_data::bcs` that purportedly allows the user/developer to switch types of boundary conditions on the different domain boundaries. On the other hand, here you are requesting that Dirichlet conditions are enforced on boundaries 0 and 2. Furthermore, the assembly method assumes Neumann conditions on boundaries 1 and 3, regardless of the values stored in `problem_data::bcs`.

Given the scope of the project, I believe it is perfectly fine to hard-code boundary conditions. If however you choose to do so, I would advise against using a variable that suggests a flexibility that just isn't there.

src/DTR_mg.cpp
```diff
@@ -0,0 +1,473 @@
...
+                   solution,
+                   right_hand_side,
+                   preconditioner);
+
+    constraints.distribute(solution);
+
+    pcout << "Time solve (" << solver_control.last_step() << " iterations)"
+          << (solver_control.last_step() < 10 ? "  " : " ") << "(CPU/wall) "
+          << time.cpu_time() << "s/" << time.wall_time() << "s\n";
+    pcout << "   Number of CG iterations:      " << solver_control.last_step()
```
michelebucelli, line 394:
```suggestion
    pcout << "   Number of GMRES iterations:      " << solver_control.last_step()
```

include/DTR_mf.hpp
```diff
@@ -0,0 +1,385 @@
...
+     *
+     * @param cycle The cycle number for which to output the results.
+     */
+    void output_results(const unsigned int cycle) const;
+
+#ifdef DEAL_II_WITH_P4EST
+    parallel::distributed::Triangulation<dim> triangulation;
+#else
+    Triangulation<dim> triangulation;
+#endif
```
michelebucelli, line 285: Is there a specific reason why you support triangulations both with and without p4est? If so, have you tested both?

michelebucelli (commented):
Thanks! I have reviewed the code and left just a few minor suggestions.

One major point I'd like to raise is that there is a good deal of repeated code between the implementations of the different solution approaches (DTR, DTR_mg, DTR_mf), e.g. for managing the mesh, the DoF handler, the output. It might be a good programming exercise to try and think of a way to reduce/eliminate these repetitions.

PR #9, opened 2024-07-08
Consegna codice progetto "Fisher-Kolmogorov equation for neurodegenerative diseases" del gruppo Bettonte, Lacagnina, Lentini

fablnt:


include/FisherKolmogorov1D.hpp
```diff
@@ -0,0 +1,161 @@
...
+  {
+  public:
+    FunctionC0()
+    {}
+
+    virtual double
+    value(const Point<dim> &p,
+          const unsigned int /*component*/ = 0) const override
+    {
+      if ((p[0] - 0) == 0)
```
michelebucelli, line 57: This assumes that there exists a mesh vertex exactly located at $x = 0$. Depending on the number and location of the elements, this may or may not be true even if the domain does include $x = 0$, and depending on how the mesh is generated it may not be true even as $h \to 0$.

This may lead to issues where the initial concentration appears to be zero everywhere, even though its continuous definition is different from zero.

To avoid this, it might be preferrable to use something like `std::abs(p[0] - 0.0) < tolerance` (which at least always works as the mesh is refined).

Also beware that equality comparisons between floating point numbers is generally ill-advised (if the numbers arise from arithmetic operations, finite precision might cause the equality comparison to be false even if it should be true in exact precision).

include/FisherKolmogorov1D.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+  FisherKolmogorov1D(const std::string  &mesh_file_name_,
+                     const unsigned int &r_,
+                     const double       &deltat_,
+                     const double       &T_,
+                     const double       &spreading_coeff_,
+                     const double       &growth_coeff_)
+    : growth_coefficient(growth_coeff_)
+    , spreading_coefficient(spreading_coeff_)
+    , c_0()
```
michelebucelli, line 72: This is actually superfluous.

src/FisherKolmogorov1D.cpp
```diff
@@ -0,0 +1,294 @@
...
+  {
+    std::cout << "Initializing the finite element space" << std::endl;
+
+    fe = std::make_unique<FE_Q<dim>>(r);
+
+    std::cout << "  Degree                     = " << fe->degree << std::endl;
+    std::cout << "  DoFs per cell              = " << fe->dofs_per_cell
+              << std::endl;
+
+    quadrature = std::make_unique<QGaussSimplex<dim>>(r + 1);
```
michelebucelli, line 28: I think this quadrature formula should be of type `QGauss`.

src/FisherKolmogorov1D.cpp
```diff
@@ -0,0 +1,294 @@
...
+void
+FisherKolmogorov1D::solve_linear_system()
+{
+  // setting for solver
+  SolverControl solver_control(
+    1000, 1e-6 * residual_vector.l2_norm()); // residual norm taken into account
+                                             // for a more reliable tolerance
+
+  // GMRES solver
+  SolverGMRES<Vector<double>> solver(solver_control);
```
michelebucelli, line 170: The linear system is actually symmetrical (the i-th and j-th shape functions appear in the same role in the assembling). Therefore, using the conjugate gradient method might be beneficial for computational efficiency.

include/BrainRegions.hpp
```diff
@@ -0,0 +1,580 @@
...
+   */
+  virtual bool
+  check_region(const Point<dim> &p) const = 0;
+};
+
+/**
+ * Derived class for Tau inclusions in the brain.
+ */
+template <int dim>
+class Tau_inclusions : public SeedingRegion<dim>
```
michelebucelli, line 39: This class and the following ones would likely be simpler/shorter if, instead of using template specialization over the physical dimension, you just used `if constexpr` statements.

include/BrainRegions.hpp
```diff
@@ -0,0 +1,580 @@
...
+ * The direction is specified by the user in the main.cpp file.
+ *
+ * @param orientation Name of the orientation.
+ * @param global_center Global center of the brain.
+ * @return Unique pointer to the direction.
+ * @throw std::invalid_argument if the orientation name is invalid.
+ */
+template <int dim>
+std::unique_ptr<Function<dim>>
+get_direction(std::string orientation, const Point<dim> &global_center)
```
michelebucelli, line 568:
```suggestion
get_direction(const std::string &orientation, const Point<dim> &global_center)
```

include/BrainRegions.hpp
```diff
@@ -0,0 +1,580 @@
...
+ * Factory function to create seeding regions.
+ * The seeding region is specified by the user in the main.cpp file.
+ *
+ * @param region Name of the seeding region.
+ * @return Unique pointer to the seeding region.
+ * @throw std::invalid_argument if the seeding region name is invalid.
+ */
+template <int dim>
+std::unique_ptr<SeedingRegion<dim>>
+getSeedingRegion(std::string region)
```
michelebucelli, line 209:
```suggestion
getSeedingRegion(const std::string &region)
```

include/BrainRegions.hpp
```diff
@@ -0,0 +1,580 @@
...
+  {
+    Vector<double> radial(dim);
+    for (unsigned int i = 0; i < dim; ++i)
+      radial[i] = p[i] - global_center[i];
+    radial /= radial.l2_norm();
+
+    values[0] = radial[0];
+    values[1] = radial[1];
+    if (dim == 3)
+      values[2] = radial[2];
```
michelebucelli, line 313: The following would be dimension independent in a more flexible way:
```suggestion
    for (unsigned int i = 0; i < dim; ++i)
      values[i] = radial[i];
```

include/BrainRegions.hpp
```diff
@@ -0,0 +1,580 @@
...
+        values[0] = radial[1] * orthogonal[2] - radial[2] * orthogonal[1];
+        values[1] = radial[2] * orthogonal[0] - radial[0] * orthogonal[2];
+        values[2] = radial[0] * orthogonal[1] - radial[1] * orthogonal[0];
+      }
+  }
+
+  virtual double
+  value(const Point<dim> &p, const unsigned int component = 0) const override
+  {
+    Vector<double> radial(dim);
```
michelebucelli, line 419: This function has a lot of duplicated code from the one above. It is probably better to just call `vector_value` and return the requested component.

include/FisherKolmogorov.hpp
```diff
@@ -0,0 +1,321 @@
...
+
+  /**
+   * Function for the initial concentration of the misfolded
+   * protein c(t = 0), depending on the seeding region of the problem.
+   */
+  class FunctionC0 : public Function<dim>
+  {
+  public:
+    FunctionC0(const std::string &seeding_region_)
+      : seeding_region(std::move(getSeedingRegion<dim>(seeding_region_)))
```
michelebucelli, line 169: I believe that you don't need to explicitly call `std::move` here, the unique pointer is moved automatically.

src/FisherKolmogorov.cpp
```diff
@@ -0,0 +1,444 @@
...
+ * tetrahedral elements and chooses the Gauss-Legendre quadrature formula to
+ * compute the required integrals. Initializes the Dof Handler, to identify and
+ * interact with the degrees of freedom located in the mesh. Initializes the
+ * grey and white matter region in order to set different parameters depending
+ * on the type of element considered.
+ * Computes the center of the domain, necessary to implement the axonal
+ * transport coefficient (parameter), and initializes the fiber orientation.
+ * Initializes the structure of the linear system, which is what the starting
+ * problem is reduced to.
+ */
```
michelebucelli, line 19: For future reference, it is customary to put documentation comments such as this one in the header file.

Documentation is mostly needed by class users, that ideally would not need to know about the internal implementation details, but only need to know the interface, and thus would typically only read the hpp file.

src/FisherKolmogorov.cpp
```diff
@@ -0,0 +1,444 @@
...
+        local_center += cell->center();
+      }
+
+    MPI_Allreduce(
+      &local_center, &global_center, dim, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+
+    global_center /= mesh.n_global_active_cells();
+
+    pcout << "Center of the domain = " << global_center << std::endl;
+  }
```
michelebucelli, line 103: This is a rather approximate way of computing the center, because it does not caaount for the size of cells. The correct way of doing it is to compute $\bar{\mathbf x} = \int_{\Omega}\mathbf{x} d\mathbf{x}$ (through an appropriate quadrature rule).

PR #10, opened 2024-07-09
Non-linear cardiac mechanics (Chen K.M., De March N., Raffaelli M.)

mattteochen:
### This PR submits the Finite Element Solver implementation for the "Non-linear cardiac mechanics" project.

Please find all the needed setup and compilation steps in the readme.

The repository was auto-formatted with `clang-format` using the `LLVM` rules.

Custom meshes (**with compatible geometry**) can be used, for more info refer to the readme.

The meshes `.geo` files that we have used in order to run our tests can be found under `mesh/`.

An offline documentation of the solver classes can be found under `src/html/index.html`.


michelebucelli:
As a general rule, it is better to avoid uploading files that are the result of compilation, in favor of uploading the scripts/makefiles/rule files needed to compile them. In this case, you should avoid uploading the generated documentation, and instead simply modify your build system to allow for something like `make doc`.

**Edit**: the large number of files seems to be causing freezes in GitHub review system, making it neary impossible to review the code. I will suspend the revision until you remove the documentation compiled files.

src/tests/new_hook/slab_cubic/SlabCubicNewHook.hpp
```diff
@@ -0,0 +1,73 @@
+/**
+ * @file SlabCubicNewHook.hpp
```
michelebucelli: The model is called *neo Hooke*, not *New Hook*.

src/tests/new_hook/slab_cubic/SlabCubicNewHook.hpp
```diff
@@ -0,0 +1,73 @@
...
+      : Base(problem_name_),
+        zero_function(dealii::Functions::ZeroFunction<dim>(dim)) {
+    initialize_param_handler(parameters_file_name_);
+    initialise_boundaries_tag();
+  }
+  /**
+   * @see Base::initialise_boundaries_tag
+   */
+  void initialise_boundaries_tag() override {
+    //  Set Newmann boundary faces
```
michelebucelli: The boundary conditions are called *Neumann*, not *Newmann*.

mattteochen:
> As a general rule, it is better to avoid uploading files that are the result of compilation, in favor of uploading the scripts/makefiles/rule files needed to compile them. In this case, you should avoid uploading the generated documentation, and instead simply modify your build system to allow for something like `make doc`.
> 
> **Edit**: the large number of files seems to be causing freezes in GitHub review system, making it neary impossible to review the code. I will suspend the revision until you remove the documentation compiled files.

We have removed the documentation from tracking, sorry for this issue. Now the PR should be fine to be reviewed.

In the next days we will integrate it inside CMake, for now it can be generated by simply running doxygen directly (if needed).

Thanks.

README.md
```diff
@@ -1,11 +1,133 @@
...
+the left ventricle when a force acts on the internal surface of the ventricle, known as the
+Endocardium. We model the acting forces as pressure without considering
+how the force is generated. Our modeling strategy follows a bottom-up approach: initially,
+we employ a simplified geometry and basic mathematical model, making assumptions about
+material properties. Subsequently, we refine the model by incorporating a more intricate
+formulation that accounts for the material properties. This testing methodology is inspired
+by the approach outlined [here](https://pubmed.ncbi.nlm.nih.gov/26807042/).
 
-Binary files must not be uploaded to the repository (including executables).
+## Constructive law: Guccione
```
michelebucelli: The stress-strain relation is called "constitutive law", not constructive.

src/lib/cardiac_mechanics/BaseSolverNeoHooke.hpp
```diff
@@ -0,0 +1,370 @@
...
+          reduction_increment_strat(reduction_factor_inc_) {}
+    /**
+     * @brief Copy operator
+     * @param other The input ConstantPressureFunction object
+     */
+    void operator=(ConstantPressureFunction &&other) {
+      pressure = other.pressure;
+      reduction_factor = other.reduction_factor;
+      reduction_increment_strat = other.reduction_increment_strat;
+    }
```
michelebucelli: I believe this operator would be default-generated, there should be no need to declare it manually. If you want to be explicit, you could say
```suggestion
    void operator=(ConstantPressureFunction &&other) = default;
```

src/lib/cardiac_mechanics/BaseSolverNeoHooke.hpp
```diff
@@ -0,0 +1,370 @@
...
+     */
+    static std::string show() {
+      return "  mu: " + std::to_string(mu) + "\n" +
+             "  lambda: " + std::to_string(lambda) + "\n";
+    }
+  };
+  /**
+   * @brief Class representing the applied pressure component
+   */
+  class ConstantPressureFunction : Function<dim> {
```
michelebucelli: There's a bit of a conflict here: the `Function<dim>` class is meant to return a `double`, while the `ConstantPressureFunction` is meant to return a `Scalar`, which just so happens to be a `double` by default, but it might be something else. If you were to declare `BaseSolverNeoHooke` with a different template argument, you would get warnings/errors/implicit conversions, none of which are good.

I know that you are asserting that `Scalar` is `double` in the constructor, but this still leaves you open to possible issues should this assertion be removed/changed in the future.

Two possible solutions to this are to use the second template argument to `Function` to make sure it returns a `Scalar` as well, or remove templatization over the scalar type altogether.

src/lib/cardiac_mechanics/BaseSolverNeoHooke.hpp
```diff
@@ -0,0 +1,370 @@
...
+     * paper for the employed modfied Newton iteration
+     */
+    double reduction_increment_strat;
+  };
+  /**
+   * @brief Determine if a given faces is at a Neumann boundary
+   * @param face The face values
+   * @return The requested query
+   */
+  bool is_face_at_neumann_boundary(const unsigned face) {
```
michelebucelli: This should be a `const` method, since it does not modify the state of the object.

src/lib/cardiac_mechanics/BaseSolverNeoHooke.hpp
```diff
@@ -0,0 +1,370 @@
...
+   * @brief Initialise boundaries tag. This must be implemented by the derived
+   * class
+   */
+  virtual void initialise_boundaries_tag() = 0;
+  /**
+   * @brief Initialise parameters handler. This must be implemented by the
+   * derived class
+   * @param file_ The input parameters file
+   */
+  virtual void initialize_param_handler(const std::string &file_) = 0;
```
michelebucelli, line 227: The fact that this method needs to be implemented by the derived classes is somewhat counterintuitive. I would think that the initialization of the parameter handler was independent of the underlying problem, while each class would have to implement their own version of `declare_parameters` and `parse_parameters` to manage its problem-specific parameters.

src/lib/cardiac_mechanics/BaseSolverNeoHooke.cpp
```diff
@@ -0,0 +1,667 @@
...
+          if (cell->face(face_number)->at_boundary() &&
+              is_face_at_neumann_boundary(
+                  cell->face(face_number)->boundary_id())) {
+            fe_values_boundary.reinit(cell, face_number);
+
+            // Loop over face quadrature points
+            for (unsigned int q = 0; q < n_face_q; ++q) {
+              // Compute deformation gradient tensor
+              const auto F =
+                  Physics::Elasticity::Kinematics::F(solution_gradient_loc[q]);
```
michelebucelli: This is not correct!

The variable `solution_gradient_loc` contains the evaluation of the displacement gradient at the quadrature nodes **within the volume of the cell** (since they have been obtained through `fe_values`). Here, you need to evaluate the gradient at the quadrature nodes **on the surface of the face**, through `fe_values_face`. The tensor you compute here is more or less unrelated to the one you want to compute, and this is likely to give rise to all sorts of issues.

michelebucelli:
There appears to be a significant amount of duplicated code between `BaseSolverNeoHooke` and `BaseSolverGuccione`. You could exploit the fact that they are solving two problems in the same form, that only differ in the constitutive law, and define a structure such as this one:
```cpp
// Abstract interface for a constitutive law.
class ConstitutiveLaw
{
  public:

  // Virtual method to compute the stress tensor given the deformation
  // gradient (possibly replacing doubles with AD types, and possibly)
  // adding input arguments for e.g. fiber orientation).
  virtual Tensor<2, dim, double>
  compute_P(const Tensor<2, dim, double> &F) = 0;
}

// Concrete NeoHooke constitutive law.
class NeoHooke : public ConstitutiveLaw
{
  public:

  // Implement the computation of the stress tensor.
  virtual Tensor<2, dim, double>
  compute_P(const Tensor<2, dim, double> &F) { /* ... */ }
}

// Concrete Guccione constitutive law.
class Guccione : public ConstitutiveLaw
{
  public:

  // Implement the computation of the stress tensor.
  virtual Tensor<2, dim, double>
  compute_P(const Tensor<2, dim, double> &F) { /* ... */ }
}

// General solver for non-linear hyperelastic mechanics.
class MechanicsSolver
{
  public:

  // ...all the stuff you need...

  protected:

  // Pointer to the ConstitutiveLaw, initialized at runtime,
  // possibly depending on the value of some argument.
  std::unique_ptr<ConstitutiveLaw> constitutive_law;
}
```

Within the assembly method, you would just call `constitutive_law->compute_P(F)` to retrieve the stress tensor.

src/lib/cardiac_mechanics/BoundariesUtility.hpp
```diff
@@ -0,0 +1,92 @@
...
+  BoundariesUtility(const std::string &dirichlet_boundaries_,
+                    const std::string &neumann_boundaries_) {
+    std::istringstream dirichlet_iss(dirichlet_boundaries_);
+    std::istringstream neumann_iss(neumann_boundaries_);
+    for (std::string token; std::getline(dirichlet_iss, token, ',');) {
+      dirichlet_boundaries.push_back(std::stoi(token));
+    }
+    for (std::string token; std::getline(neumann_iss, token, ',');) {
+      neumann_boundaries.push_back(std::stoi(token));
+    }
```
michelebucelli, line 42: As it happens, `dealii` offers some nifty utility functions for this purpose. You should look into `Utilities::split_string_list` and the namespace `Utilities` as a whole.

src/lib/cardiac_mechanics/BoundariesUtility.hpp
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * @brief Get the Neumann boundary tags.
+   * @return A constant reference to a vector containing Neumann boundary tags.
+   */
+  const std::vector<unsigned int> &get_neumann_boundaries_tags() const {
+    return neumann_boundaries;
+  }
+
+protected:
+  std::vector<unsigned int>
```
michelebucelli: `dealii` uses the alias `types::boundary_id` to represent unsigned integers for boundary labels. Although I'm pretty sure it falls back onto `unsigned int`, using the alias is more expressive in that it clarifies the purpose of the variable.

Similar aliases exist for other situations (e.g. `types::global_dof_index`, `types::material_id`). As a general rule, it is a good idea to introduce and use this sort of aliases to make the code self-documenting.

src/lib/cardiac_mechanics/LinearSolverUtility.hpp
```diff
@@ -0,0 +1,320 @@
...
+   * @param rhs The right-hand side vector.
+   * @param preconditioner The preconditioner object.
+   */
+  template <typename Solver, typename Matrix, typename Solution,
+            typename Vector, typename Preconditioner>
+  void solve(std::unique_ptr<Solver> const &solver, Matrix const &matrix,
+             Solution &solution, Vector const &rhs,
+             std::unique_ptr<Preconditioner> const &preconditioner) {
+    // We have to cast the solver pointer to the derived type as the dealii base
+    // solver class has not the solve method
```
michelebucelli, line 220: Since, unfortunately, the base class does not expose the interface you need, it might make more sense to work directly with the concrete classes, either instantiating the solver on the fly (instead of passing it as argument to the `solve` method - see also my other comment) or making this method a template over the concrete solver type.

src/lib/cardiac_mechanics/LinearSolverUtility.hpp
```diff
@@ -0,0 +1,320 @@
...
+#include <ostream>
+#include <string>
+
+/**
+ * @class LinearSolverUtility
+ * @brief Utility class for configuring linear solvers.
+ * @tparam Scalar The scalar type used in the linear solver configuration
+ * (default is double).
+ */
+template <typename Scalar = double> class LinearSolverUtility {
```
michelebucelli, line 26: If I understand correctly, this class does not manage the solver itself. Instead, it provides methods to initialize a solver and a preconditioner and to use them, but their actual storing is demanded to the user of this class (i.e. the class that makes use of it).

Therefore, from a user's perspective, if I want to use this class, I would have to declare an instance of `LinearSolverUtility`, an instance of linear solver and an instance of preconditioner. Furthermore, the latter two would need to be compliant with the expectations of `LinearSolverUtility` (that is, they would need to be `unique_ptr`s). This is putting a lot of work (in relative terms) on the user.

A better alternative would be to directly declare the solver and preconditioner within `LinearSolverUtility`, and hide them from the user. The user would only have to specify the relevant settings (which type of solver and of preconditioner, tolerances and so on).

Furthermore, the `LinearSolverUtility` class might also take care of declaring and parsing the parameters associated to the linear solver and preconditioner, so that the user would not even have to worry about that.

As a general rule, wrappers such as this one should be designed in such a way that the user interface is simplified with respect to working without the wrapper. This simplification often comes in terms of reducing the amount of boilerplate code and of code in general needed to achieve a certain feature.

src/lib/cardiac_mechanics/SolverException.hpp
```diff
@@ -0,0 +1,37 @@
...
+#define SOLVER_EXCEPTION_HPP
+
+#include <deal.II/base/exceptions.h>
+#include <string>
+
+/**
+ * @class SolverException
+ * @brief Utility class for defining the base class for runtime exceptions.
+ */
+class SolverException : public dealii::ExceptionBase {
```
michelebucelli, line 16: I believe that the same functionality of this class is achieved through `dealii::ExceptionMsg`.

src/lib/cardiac_mechanics/NewtonSolverUtility.hpp
```diff
@@ -0,0 +1,66 @@
...
+#define NEWTON_SOLVER_UTILITY_HPP
+
+#include <iostream>
+
+/**
+ * @class NewtonSolverUtility
+ * @brief Utility class for configuring Newton solvers.
+ * @tparam Scalar The scalar type used in the solver configuration.
+ */
+template <typename Scalar> class NewtonSolverUtility {
```
michelebucelli, line 17: Similarly to the linear solver utility, this class is a bit of a missed opportunity to wrap the Newton loop within an external tool, to spare the user from having to reimplement it every time.

src/lib/cardiac_mechanics/BaseSolverNeoHooke.cpp
mattteochen: Yes, this is wrong. Thank you. We did spot the same issue for the other constitutive law but we missed checking this on neo Hooke.

src/lib/cardiac_mechanics/BaseSolverNeoHooke.hpp
mattteochen: Yes, that makes sense.

FYI the assert on the `double` type was put to accommodate trillinos data structures, which (if I am not wrong) supports only doubles

src/lib/cardiac_mechanics/BoundariesUtility.hpp
mattteochen: Thanks for the nit! They make code cleaner for sure.

src/lib/cardiac_mechanics/BaseSolverNeoHooke.hpp
mattteochen: I agree that this could be made cleaner and better, the long story short for this choice is that we needed to align some std::cout, hence only a visual accommodation, probability not worth breaking good programming paradigms for this cause.

src/lib/cardiac_mechanics/LinearSolverUtility.hpp
mattteochen: I really appreciate this observation, thanks!

src/lib/cardiac_mechanics/SolverException.hpp
mattteochen: Yes, 100%. In the beginning, I only wanted to give a clear and fast understanding of the issue by using a custom class name. Now it might or might not be displayed during exceptions making this choice useful or not. Will check it.

mattteochen:
@michelebucelli Thank you for the review. We fixed the critical issue and some minors. Will complete other optimizations and enhancements soon.

PR #11, opened 2024-07-09
Navier Stokes equations - project 3 (Tortorella Mara, Marzagora Bice, Selis Riccardo)

BiceMar:
Consegna progetto


src/main.cpp
```diff
@@ -0,0 +1,185 @@
+#include <getopt.h>
+#include "NavierStokes.hpp"
+#include "NavierStokes.cpp"
+
+void print_usage() {
+    std::cout << "Usage: ./navier_stokes_solver\n --mesh <mesh_file> : mesh-0.1, mesh-0.05, mesh-0.025, mesh-0.0125\n --degree_velocity <degree>\n --degree_pressure <degree>\n --T <total_time>\n --deltat <time_step>\n  --theta <theta>\n --nu <viscosity>\n --p_out <pressure>\n --rho <density>\n --velocity_case_type <case_type>\n --vel <velocity>\n --prec <preconditioner>: 0 for block-diagonal, 1 for SIMPLE, 2 for aSIMPLE, 3 for Yosida\n --dim <dim>: 2 for 2D, 3 for 3D\n --use_skew <use skew symmetric representation>: 0 for skew, 1 for actual non-linear term\n";
```
michelebucelli, line 6: For the sake of readability, it is usually preferrable to avoid long lines such as this one. You can break this string into multiple lines as follows:
```cpp
std::cout << "this is the beginning of a long string "
                      "which I split over a few lines so that "
                      "it does not seem that long"
          << std::endl;
```
or by concatenating multiple `<<` operators.

src/main.cpp
```diff
@@ -0,0 +1,185 @@
...
+        {"dim", required_argument, 0, 'i'},
+        {"use_skew", required_argument, 0, 'u'},
+        {"help", no_argument, 0, 'h'},
+        {0, 0, 0, 0}
+    };
+
+    bool show_help = false;
+    int option_index = 0;
+    int c;
+    while ((c = getopt_long(argc, argv, "m:V:P:T:d:t:n:p:r:c:v:e:i:u:h", long_options, &option_index)) != -1) {
```
michelebucelli, line 55: `getopt` is a perfectly valid tool, although you might want to consider more moden, high-level tools such as `clipp`.

src/NavierStokes.hpp
```diff
@@ -0,0 +1,743 @@
...
+    }
+
+  protected:
+    const double g = 0.0;
+  };
+
+  class InletVelocity : public Function<dim>
+  {
+  public:
+      InletVelocity(int case_type, double vel)   // Default to case 1 if not specified
```
michelebucelli, line 75: The comment is misleading. This constructor does not have a default value for the `case_type` argument. Be careful to avoid incorrect documentation, which is worse than no documentation at all.

src/NavierStokes.hpp
```diff
@@ -0,0 +1,743 @@
...
+  FunctionH function_h;
+
+  // Vector of all the drag coefficients
+  std::vector<double> drag_coefficients; 
+  
+  // Vector of all the drag coefficients
+  std::vector<double> lift_coefficients;
+
+  // Retrive Drag/Lift coefficient multiplicative constant
+  double get_drag_lift_multiplicative_const(double t) {
```
michelebucelli, line 673:
```suggestion
  double get_drag_lift_multiplicative_const(double t) const {
```
If a method does not require to change the state of a class, always declare it `const`, so that it can also be called on `const` instances of (or `const` references to) that class.

src/NavierStokes.hpp
```diff
@@ -0,0 +1,743 @@
...
+    double multiplicative_const;
+
+    if constexpr(dim == 2) multiplicative_const = 2.0 / (rho * inlet_velocity.mean_velocity_value(t) * inlet_velocity.mean_velocity_value(t) * cylinder_height); 
+    if constexpr(dim == 3) multiplicative_const = 2.0 / (rho * inlet_velocity.mean_velocity_value(t) * inlet_velocity.mean_velocity_value(t) * cylinder_diameter * cylinder_height); 
+    
+    return multiplicative_const; 
+  }
+
+  // Retrive Reynolds number
+  double get_reynolds_number(double t) {
```
michelebucelli, line 683:
```suggestion
  double get_reynolds_number(double t) const {
```

src/NavierStokes.cpp
```diff
@@ -0,0 +1,758 @@
...
+                  fe_values.JxW(q);
+            }
+        }
+        else{
+          for (unsigned int j = 0; j < dofs_per_cell; ++j)
+                {
+                  // Semi-implicit treatment of the non linear term
+                  non_linear_contribution(i, j) += velocity_loc[q] * fe_values[velocity].gradient(j, q) *
+                                        fe_values[velocity].value(i, q) *
+                                        fe_values.JxW(q);
```
michelebucelli, line 359: I believe this term is not entirely correct. This corresponds to $(\mathbf{u}^n \cdot \nabla) \mathbf u \cdot \mathbf v$. In Einstein notation (repeated indices imply summation), this becomes $((\mathbf{u}^n \cdot \nabla)\mathbf u)_i \mathbf v_i = (\mathbf{u}^n \cdot \nabla) \mathbf u_i \mathbf v_i = \left( \mathbf{u}^n_j \frac{\partial}{\partial\mathbf{x}_j}\right) \mathbf{u}_i \mathbf{v}_i = \mathbf{u}^n_j \frac{\partial\mathbf u_i}{\mathbf x_j} \mathbf v_i = \mathbf u^n_j (\nabla\mathbf u)\_{ij} \mathbf v_i = (\nabla\mathbf u\mathbf u^n)_i \mathbf v_i = (\nabla\mathbf u \mathbf u^n)\cdot\mathbf v$.
Therefore, I believe this should be implemented as
```suggestion
non_linear_contribution(i, j) +=
  scalar_product(
    fe_values[velocity].gradient(j, q) * velocity_loc[q],
    fe_values[velocity].value(i, q)) * 
  fe_values.JxW(q);
```
This might explain the unusual behavior observed in your velocity field (see also my comments to the report).

PR #13, opened 2024-08-30
Bergamaschi_DiRaimondo_Fonnesu_Project

Fonzy-01:


FK_Project/src/Fisher1D.hpp
```diff
@@ -0,0 +1,215 @@
...
+    class FunctionAlpha : public Function<dim> 
+    {
+    public : 
+        virtual double 
+        value(const Point<dim> & /*p*/,
+            const unsigned int /*component*/ = 0) const /*override*/
+        {   
+            return 2.0;
+        }
+    };
```
michelebucelli, line 63: For future reference, you might want to look into the `dealii::Functions::ConstantFunction` class.

FK_Project/src/Fisher1D.hpp
```diff
@@ -0,0 +1,215 @@
...
+    {
+    public:
+        virtual double
+        value(const Point<dim> &p,
+            const unsigned int /*component*/ = 0) const /*override*/
+        {
+            //Point<dim> origin;
+            Point<dim> starting_point(0.0);
+            double max_value = 0.1;
+            double std_dev = 0.0333;
```
michelebucelli, line 89:
```suggestion
            const Point<dim> starting_point(0.0);
            constexpr double max_value = 0.1;
            constexpr double std_dev = 0.0333;
```
Declaring variables as `const` or `constexpr` greatly facilitates compiler optimization (e.g. the compiler would likely precompute `2 * std_dev * std_dev` if `std_dev` was declared as `constexpr`, and it might do so even if it was declared as `const`). Additionally, these could more appropriately be defined as members of the class, rather than local variables to the `value` method.

FK_Project/src/Fisher1D.cpp
```diff
@@ -0,0 +1,302 @@
...
+          cell_matrix(i, j) += fe_values.shape_value(i, q) *
+                              fe_values.shape_value(j, q) / deltat *
+                              fe_values.JxW(q);
+
+          // First term of the stiffness matrix
+          cell_matrix(i, j) += D_loc * fe_values.shape_grad(j, q) *
+                              fe_values.shape_grad(i, q) *
+                              fe_values.JxW(q);
+
+          // Second term of the stiffness matrix
```
michelebucelli, line 145: Strictly speaking, the "stiffness" matrix is the one related to stiffness-like terms in the PDE (i.e. the diffusion-like grad-grad terms, which are formally similar to the elastic stiffness terms in solid mechanics equations).

The terminology arises from the original application of the finite element method to structural mechanics (the mass matrix is associated to inertia, the stiffness matrix is associated to elasticity, ...).

FK_Project/src/Fisher_Kolmogorov.hpp
```diff
@@ -0,0 +1,252 @@
...
+            // Here should go the values for the d_axn
+            // Example: values(1,1) += 3.0;
+        }
+
+        virtual double 
+        value(const Point<dim> &/*p*/, const unsigned int component1 = 0, const unsigned int component2 = 1)  const /*override*/ 
+        {
+            return (component1 == component2) ? 1.0 : 0.0;
+        }
+    };
```
michelebucelli, line 85: Mathematically speaking, $D$ is a tensor-valued field, i.e. a function of space that returns a rank-2 tensor. The best representation you can give in code is that of a `Tensor<2,dim>`-valued function. `deal.ii` has a `TensorFunction<rank, dim>` class for this purpose. Alternatively, you might consider using the `Function<dim>` class with `dim*dim` components (mapping the matrix elements into a linear ordering).

Either way, you should **never** change the interface of a class you inherit from. For instance, the `Function<dim>` class has a method `double value(const Point<dim> &, const unsigned int)` that evaluates the function. Here you redefine your own `value` method, that has a different signature (different arguments) and therefore does **not** override `Function<dim>::value`, but rather it overloads it (which, I assume, is the reason why you commented out the `override` keyword).

Now, consider the following situation. Some user of your code (possibly your future selves) wants to interpolate the value of this function on all nodes of a mesh, for some reason. To do so, they call `VectorTools::interpolate`: after all, this object is a `Function<dim>`, it is reasonable to expect it to behave as such (that is the meaning of class inheritance), and so it is reasonable to believe I can feed it to any function that takes a `Function<dim>` in input.

However, `VectorTools::interpolate` calls `value(const Point<dim>&, const unsigned int)`, that is the version of the method present in the interface of `Function<dim>`. I'm not sure what the `Function<dim>::value` method does by default, but I'm pretty sure it does not do the same as your overloaded version. Maybe it throws an exception, maybe it returns zero without giving any warning. In the latter case, you have a very hard to find bug (especially if the user of the code does not know or remember the design choices made in the implementation of `FunctionD`).

tl;dr: never depart from the expected interface of a parent class unless you have a very very very strong reason to do so and you trust your software design skills very very very much. Otherwise, you should either not inherit from that class, or change your code to work with its established interface.

PR #14, opened 2024-08-30
NMPDE project submission

tommasotrabacchinpolimi:
NMPDE project submission by Tommaso Trabacchin, Melanie Tonarelli, Sabrina Cesaroni


src/utils.hpp
```diff
@@ -0,0 +1,132 @@
...
+// N is the number of gating variables, 3 in our case
+template<int N>
+struct GatingVariables {
+    double& get(int k) {
+        //assert(k < N);
+        return var[k];
+    }
+private:
+    double var[N];
+};
```
michelebucelli, line 43: Why not just aliasing the `std::array` class?
```cpp
template<int N>
using GatingVariables = std::array<double, N>;
```
This way, you get all the niceness of the standard library, plus the synctactic sugar of the meaningful name.

src/utils.hpp
```diff
@@ -0,0 +1,132 @@
...
+        Tensor<2, dim> sigma = unit_symmetric_tensor<dim>();
+        //return sigma * 1.171e2;
+        sigma[0][0] = sigma_il * sigma_el / (sigma_il + sigma_el);
+        sigma[1][1] = sigma_it * sigma_et / (sigma_it + sigma_et);
+        sigma[2][2] = sigma_it * sigma_et / (sigma_it + sigma_et);
+        return sigma;
+    }
+
+private:
+    double sigma_il = 0.17; /*S/m*/
```
michelebucelli, line 63:
```suggestion
    const double sigma_il = 0.17; /*S/m*/
```
or even better
```suggestion
    static constexpr double sigma_il = 0.17; /*S/m*/
```
(and same for the others).

src/utils.hpp
```diff
@@ -0,0 +1,132 @@
...
+    double var[N];
+};
+
+// tissue conductivity tensor
+template <int dim>
+class D : public TensorFunction<2, dim>
+{
+public:
+    D() : TensorFunction<2, dim>() {}
+    void value_list(const std::vector<Point<dim>> &/*points*/, std::vector<Tensor<2, dim>> & /*values*/) const override {} // do not use it
```
michelebucelli, line 51: If a method is not meant to be used, but you still need to implement it (e.g. due to it being in the interface of a purely virtual class), it might be a good idea to have it throw an exception if it is used.

src/utils.hpp
```diff
@@ -0,0 +1,132 @@
...
+class Iapp : public Function<dim>
+{
+public:
+    virtual double
+    value(const Point<dim> & p, const unsigned int /*component*/ = 0) const override
+    {
+        if(p[0] <= 1.5e-3 && p[1] <= 1.5e-3 && p[2] <= 1.5e-3 /*millimeters*/ && this->get_time() <= 0.002 /* milliseconds */) {
+            return 50000;//;6.3 * 1000;/* nA/(mm)^3 */;
+        }
+        return 0.0;
```
michelebucelli, line 83:
```suggestion
        }
        else {
          return 0.0;
        }
```
This is clearly equivalent. However, the `else` branch clarifies the intent more explicitly.

src/utils.hpp
```diff
@@ -0,0 +1,132 @@
...
+// initial condition
+template <int dim>
+class U_0 : public Function<dim> {
+public:
+    virtual double
+    value(const Point<dim> & /*p*/, const unsigned int /*component*/ = 0) const override
+    {
+        return -0.08523; /*V*/
+    }
+};
```
michelebucelli, line 96: For these constant-valued functions, `deal.II` offers the convenient `ConstantFunction` class.

src/main.cpp
```diff
@@ -0,0 +1,122 @@
...
+            os =  std::stoi(argv[i + 1]);
+            ++i;
+        }
+        else {
+            if(mpi_rank == 0){
+                std::cout << "Wrong parameters!" << std::endl;
+                return 1;
+            }
+        }
+    }
```
michelebucelli, line 78: For future reference:
1. command-line arguments are more easily managed through dedicated high-level libraries. One such library is clipp, but there's many others;
2. you might want to look into `deal.II`'s facilities for reading parameters from external, human-readable files (see the documentation for the class `ParameterHandler`).

src/IonicModels/IonicModel.hpp
```diff
@@ -0,0 +1,25 @@
+#ifndef IONIC_MODEL_H
+#define IONIC_MODEL_H
+#include <array>
+#include "../utils.hpp"
+
+// Abstract class representing the ionic model, from which the BuenoOrovioModel class inherits
+template<int N_ion>
```
michelebucelli, line 7: I don't think that templating over the number of ionic variables is a sound design choice, and it would probably break as soon as you have more than one ionic model (which however is the reason for having an abstract `IonicModel` class in the first place).

Think of this situation: you have a model A, that has three variables, and a model B, that has 5. In order to be able to seamlessly switch between the two, you could think of using polymorphism and access eitheer through their common base class. The problem is that with current implementation the base class of `A` would be `Ionic<3>` and the base class of `B` would be `Ionic<5>`, and since their template arguments are different these are two different classes. Polymorphism won't work here, defying the purpose of the base class altogether.

A structure that would work better is the following:
```cpp
class IonicModel
{
public:
  // ...public interface methods...
protected:
  // Protected constructor so only derived classes can call it.
  IonicModel(const unsigned int &n_vars_)
    : n_vars(n_vars) // ...maybe more stuff...
  {}

  // Number of variables.
  unsigned int n_vars;
}

class A : public IonicModel
{
public:
  // Constructor.
  A()
    : IonicModel(3)
  {}
}

class B : public IonicModel
{
public:
  // Constructor.
  B()
    : IonicModel(5)
  {}
}
```
Now both classes inherit from `IonicModel`, and polymorphism works. Granted, you may lose a bit of efficiency due to the number variables now being unknown at compile time. However, that's to some extent the price to pay for this level of abstraction.

src/IonicModels/IonicModel.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+    virtual double ionic_current(const double u, GatingVariables<N_ion>& vars);
+
+    virtual std::tuple<double, double> getExpansionCoefficients(int index, double u);
+
+    virtual double get_FI(double u, GatingVariables<N_ion>& vars) = 0;
+
+    virtual double get_SO(double u, GatingVariables<N_ion>& vars) = 0;
+
+    virtual double get_SI(double u, GatingVariables<N_ion>& vars) = 0;
```
michelebucelli, line 21: These three appear to be specific to the `BuenoOrovio` model, and would not make sense for most other ionic models.

src/IonicModels/IonicModel.hpp
michelebucelli, line 1: As detailed in the other comments, this class seems to serve the purpose of being an abstract `IonicModel` from which other ionic models can derive, and that can provide a certain level of polymorphism. However, it fails at that purpose, for the reasons outlined in the other comments.

I think this is a good example of "unnecessary abstraction" - you only had to implement one ionic model, and so it was not evident to you which features needed to be put in the concrete class and which in the abstract class. You could have obtained a semantically equivalent program by just defining a `BuenoOrovio` class, and that program would have had basically zero code repetition or other redundancies.

In future situations, it may be a better strategy to start from the most concrete implementation, proceed bottom-up, and only introduce abstraction as necessary.

For instance: I first implement the `BuenoOrovio` model in its own class (no parent virtual class yet). After a while, I notice I also need to implement the `AlievPanfilov` ionic model. Only at that point, I implement the abstract class `IonicModel`, in which I put all the features that are common to `BuenoOrovio` and `AlievPanfilov`, having in mind the requirements of both. Of course this is not a one-fits-all strategy, but it's something useful to keep in mind, especially if you're not sure about the future requirements you might encounter.

src/IonicModels/BuenoOrovioModel/BOModel.hpp
```diff
@@ -0,0 +1,300 @@
...
+        }
+    }
+
+    // V = 85.7u − 84
+    double getAdimensionalU(double v){
+        return (1000.0 * v + 84.0) / 85.7;
+    }
+
+
+    double getDerivative(int index, double u, GatingVariables<N_ion> vars) override
```
michelebucelli, line 151:
```suggestion
    double getDerivative(const int &index,
                         const double &u,
                         const GatingVariables<N_ion> &vars) const override
```
#### 1. Methods should **always** be declared `const` if they don't need to modify the state of the object.

Also beware that the "state" should be interpreted in the sense of *semantics*, not *syntax*. In other words, the state is made up of what conceptually represents the state of the object, and not necessarily of all of its members. In other other words: you should decide *in advance* and *by design* whether a method is supposed to be *const* or not.

In this case, this method is supposed to compute the state derivative given the state, and there's no reason why this operation should modify the state itself. Therefore, the method should be `const`.

#### 2. Arguments of non-basic types should **always** be passed as const references.

Here you're doing unnecessary copies of the `GatingVariables` object. Whether basic types should be passed by const reference or not is partially up to personal preference. I think it is better to do so, since it highlights the fact that they are indeed input arguments (they are not copied and not modified, but just used in the calculations that the function does).

Notice that you would then need to modify line 153 as well to something like
```cpp
const double u_adim = getAdimensionalU(u);
```
This modification would actually be an improvement, in my opinion, providing a little bit of extra self-documentation to the code.

These comments apply elsewhere in your code. Also notice that while these might seem minor issues for relatively small programs, they build up to very large issues as programs become complicated, and it is therefore crucial to internalize these best practices as soon as possible.

src/MonodomainSolver/FESolver.hpp
```diff
@@ -0,0 +1,170 @@
+#ifndef FE_SOLVER_H
+#define FE_SOLVER_H
+
+#include "../utils.hpp"
+
+using namespace dealii;
+
+class FESolver {
+    // Physical dimension (1D, 2D, 3D)
+    static constexpr int dim = 3;
```
michelebucelli, line 10: I notice you define this `dim` variable in all classes. It would seem natural that there is an underlying assumption that all interacting classes share the same value of `dim`, although this assumption is never verified in practice.

I think this is one of the rare situations where a global variable would work best (wrapped in a dedicated namespace just to be sure).

src/Couplers/ICICoupler.hpp
```diff
@@ -0,0 +1,125 @@
...
+    // In this case the solveOde method is identical to the one in SVICoupler.
+    // Also in this case the gating variables are defined on dofs.
+    void solveOde(Solver<N_ion>& solver) override {
+        std::shared_ptr<IonicModel<N_ion>> ionicModel = solver.getIonicModel();
+        std::vector<VectorView<TrilinosWrappers::MPI::Vector>> gate_vars_views;
+        for(int i = 0; i < N_ion; i++) {
+            auto [first, last] = gate_vars_owned[i].local_range();
+            gate_vars_views.emplace_back(gate_vars_owned[i], first, last - first);
+        }
+        auto [first, last] = solver.getFESolutionOwned().local_range();
```
michelebucelli, line 29: Beware that this works as long as the partitioning of the vector is contiguous (i.e. rank 0 owns the first n_0 elements, rank 1 the second n_1, and so on and so forth). Actually, Trilinos does **not** make this assumption. Therefore, it might in principle be possible that this fails in quite a hard-to-debug way.

In this case, it would probably be best to work with the `IndexSet` of owned DoFs (the `IndexSet` class is designed basically for this purpose).

This also applies elsewhere.

PR #15, opened 2024-08-30
NMPDE project submission 

yektaKamane:
Dear Professor, 

We have submitted the code for grading.

Team members: Kamaneh, Raganato, Terrana


.DS_Store
michelebucelli, line 1: For future reference: these `.DS_Store` files should not be committed to the repository, as they are system-specific temporary files. If members of your team happen to use MacOS, it is usually a good idea to add these files to the `.gitignore` file of the repository.

It is also a good idea to always double-check what you commit (e.g. using the verbose flag for `git add`).

src/main.cpp
```diff
@@ -0,0 +1,72 @@
...
+  // Read the optional parameter file name or use a default value
+  std::string parameter_file;
+  if (argc >= 4)
+  {
+    parameter_file = argv[3];
+  }
+  else
+  {
+    parameter_file = "../input/test1.prm"; // Default parameter file
+  }
```
michelebucelli, line 48: For future reference, there's quite a few small libraries to manage command line arguments. One that I find quite nice and modern is `clipp` - it should be on GitHub.

src/Fisher_Kolmogorov_solver.hpp
```diff
@@ -0,0 +1,424 @@
...
+class FisherKol
+{
+public:
+
+  // Function of the fiber field
+  class FunctionN
+  {
+  public:
+
+    FunctionN(const FisherKol<dim> &fisher_kol) : fisher_kol(fisher_kol) {}
```
michelebucelli, line 62: While I understand the purpose of forwarding the `FisherKol` instance to the `FunctionN`, this unnecessarily creates a two-way coupling between the two classes, which is generally a bad idea.

It would probably be best if the `FunctionN` class simply took in input the parameters it needs (in this case, the coefficients `d_ext` and `d_axn` and the normal vector, ideally already formatted as a `Tensor<1, dim>`.

src/Fisher_Kolmogorov_solver.hpp
```diff
@@ -0,0 +1,424 @@
...
+        n_vector[index] = std::stod(item);
+        index++;
+      } 
+
+      // making sure the vector is normalized
+      const double norm = n_vector.norm();
+      if (norm != 0.0) n_vector /= norm;
+
+      // computing the D_matrix
+      Tensor<2, dim> identity_matrix = unit_symmetric_tensor<dim>();
```
michelebucelli, line 89:
```suggestion
      const Tensor<2, dim> identity_matrix = unit_symmetric_tensor<dim>();
```

src/Fisher_Kolmogorov_solver.hpp
```diff
@@ -0,0 +1,424 @@
...
+      std::string list_string = fisher_kol.parameters.get("normal_vector");
+      std::stringstream ss(list_string);
+      std::string item;
+
+      unsigned int index = 0;
+      while (std::getline(ss, item, ','))
+      {
+        n_vector[index] = std::stod(item);
+        index++;
+      }
```
michelebucelli, line 82: `deal.II` offers a convenient function `split_string_list` for this exact purpose.

src/Fisher_Kolmogorov_solver.hpp
```diff
@@ -0,0 +1,424 @@
...
+    virtual double
+    value(const Point<dim> & p,
+          const unsigned int /*component*/ = 0) const override
+    {
+      if (dim == 2) {
+        return std::cos(M_PI * p[0]) * std::cos(M_PI * p[1]) + 2;
+      }
+      if (dim == 3) {
+        return 0.0;
+      }
```
michelebucelli, line 160:
```suggestion
      static_assert(dim == 2 || dim == 3,
                    "Dimensions other than 2 or 3 are not supported");
                    
      if constexpr (dim == 2) {
        return std::cos(M_PI * p[0]) * std::cos(M_PI * p[1]) + 2;
      }
      else // if constexpr (dim == 3) {
        return 0.0;
      }
```

1. If your code assumes that something must be true (in this case, `dim` is either 2 or 3) it is generally a good habit to use assertions to verify it. Some of these assertions may be redundant (e.g. maybe you already check somewhere else that `dim` is either 2 or 3), but this redundancy can often help finding bugs. You can do this in several ways:
    1. static assertions (for compile-time dependent conditions). This adds zero computational overhead;
    2. debug assertions (e.g. through the `Assert` macro in `deal.II`). This adds a bit of computational overhead but only in debug mode, and can be used for expensive run-time checks
    3. release assertions (e.g. through the `AssertThrow` macro in `deal.II`). This does add computational overhead even in release mode, and should be used for errors depending on user-provided parameters or exceptions that need handling.

2. Using `if constexpr` allows the compiler to optimize away entire branches of the code (which it is probably figuring out on its own here, but it's still better to be explicit).

src/Fisher_Kolmogorov_solver.hpp
```diff
@@ -0,0 +1,424 @@
...
+      parameters.declare_entry("normal_vector", "0.0, 0.0, 0.0", Patterns::List(Patterns::Double(), 3, 3), "dummy");
+
+      parameters.declare_entry("T", "0.0", Patterns::Double(), "dummy");
+      parameters.declare_entry("deltat", "0.0", Patterns::Double(), "dummy");
+      parameters.declare_entry("degree", "0", Patterns::Integer(), "dummy");
+
+      parameters.declare_entry("radius", "10.0", Patterns::Double(), "dummy");
+      parameters.declare_entry("center_coords", "0.0, 0.0, 0.0", Patterns::List(Patterns::Double(), 3, 3), "dummy");
+
+      parameters.parse_input(prm_file);
```
michelebucelli, line 295: For the sake of efficiency, it is usually a good idea to parse the input parameters once and for all and store them in a member variable, since string-to-other conversions are quite costly if they are done a lot of times. Something like:
```cpp
// Locally declared parameter handler.
ParameterHandler parameters;

parameters.declare_entry("coef_alpha", "1.0", Patterns::Double(), "dummy");
// ...more declarations...

parameters.parse_input();

// Extract the values of the parameters:
prm_coef_alpha = parameters.get_double("coef_alpha");
// ...more get_* calls...
```

src/Fisher_Kolmogorov_solver.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+      for (unsigned int q = 0; q < n_q; ++q)
+        {
+          Tensor<2, dim> D_matrix = fiber.value(fe_values.quadrature_point(q));
+
+          double f_loc = 0.0;
+          if (convergence_test)
+            f_loc = forcing_term_conv.value(fe_values.quadrature_point(q));
+          else 
+            f_loc = forcing_term_conv.value(fe_values.quadrature_point(q));
```
michelebucelli, line 156: It would seem that both branches are doing the same...
```suggestion
          const double f_loc = forcing_term_conv.value(fe_values.quadrature_point(q));
```

src/Fisher_Kolmogorov_solver.cpp
```diff
@@ -0,0 +1,404 @@
...
+      cell_matrix   = 0.0;
+      cell_residual = 0.0;
+
+      fe_values.get_function_values(solution, solution_loc);
+      fe_values.get_function_gradients(solution, solution_gradient_loc);
+      fe_values.get_function_values(solution_old, solution_old_loc);
+
+      for (unsigned int q = 0; q < n_q; ++q)
+        {
+          Tensor<2, dim> D_matrix = fiber.value(fe_values.quadrature_point(q));
```
michelebucelli, line 150:
```suggestion
          const Tensor<2, dim> D_matrix = fiber.value(fe_values.quadrature_point(q));
```

src/Fisher_Kolmogorov_solver.hpp
```diff
@@ -0,0 +1,424 @@
...
+        index++;
+      } 
+
+      // making sure the vector is normalized
+      const double norm = n_vector.norm();
+      if (norm != 0.0) n_vector /= norm;
+
+      // computing the D_matrix
+      Tensor<2, dim> identity_matrix = unit_symmetric_tensor<dim>();
+      values = d_ext * identity_matrix + d_axn * outer_product(n_vector, n_vector);
```
michelebucelli, line 90: This is a bit of an ad-hoc optimization, but it's a useful trick to keep in mind.

The value that you return here is always the same (since the diffusion coefficients are constant and so is the axonal direction). Therefore, it is way more efficient to compute this once and for all in the constructor, and then simply return the precomputed value.

src/Fisher_Kolmogorov_solver.cpp
```diff
@@ -0,0 +1,404 @@
...
+                  fe_values_boundary.reinit(cell, face_number);
+
+                  for (unsigned int q = 0; q < quadrature_boundary->size(); ++q)
+                    for (unsigned int i = 0; i < dofs_per_cell; ++i)
+                      cell_residual(i) -= function_h.value(fe_values_boundary.quadrature_point(q)) * // h(xq)
+                                          fe_values_boundary.shape_value(i, q) *                     // v(xq)
+                                          fe_values_boundary.JxW(q);                                 // Jq wq
+                }
+            }
+        }
```
michelebucelli, line 233: This is technically correct. However, it would be quite unusual in this case to have a Neumann condition different from zero, given the physical meaning of the boundaries. Therefore, here you are performing several computations whose effect is that of adding a bunch of zeros to `cell_residual`.

I think you could just do without this section altogether.

src/FisherKolmogorov1D_convergence.cpp
michelebucelli, line 1: Isn't this the same as `FisherKolmogorov_solver.cpp`, except that it instantiates the 1D version?

PR #16, opened 2025-01-10
Project submission

DanieleGazzola:


include/functions.h
```diff
@@ -0,0 +1,69 @@
...
+
+template <int dim>
+class BetaFunction : public Function<dim>
+{
+public:
+    virtual Tensor<1, dim, double> gradient(const Point<dim> &p, const unsigned int component = 0) const override;
+
+    template <typename number>
+    Tensor<1, dim, number> gradient(const Point<dim, number> &p, const unsigned int component = 0) const;
+};
```
michelebucelli, line 30: If I interpret this correctly, you are using the `gradient` method because you want to evaluate a rank-1-tensor-valued function. This is not the way the `Function` class is intended to be used.

The standard way to evaluate a vector-valued function is to use the `component` argument to the `value` method to select the component to evaluate, AND to implement an override of `vector_value`. As an alternative, in some situations you can use the class `TensorFunction`.

src/functions.cpp
```diff
@@ -0,0 +1,109 @@
...
+double MuFunction<dim>::value(const Point<dim> & p, const unsigned int component) const
+{
+    return value<double>(p, component);
+}
+
+template <int dim>
+template <typename number>
+number MuFunction<dim>::value(const Point<dim, number> & /*p*/, const unsigned int /*component*/) const
+{
+    return 0.1; // -∇·(µ∇u)
```
michelebucelli, line 15: The comment is misleading, since the returned value is not $-\nabla\cdot(\mu \nabla u)$ but only $\mu$.

Always remember that wrong/misleading documentation is worse than no documentation. This is a small instance of this problem, but it's important to always adhere to good practices, in big and small things.

The same applies to other functions below.

src/setup.cpp
```diff
@@ -0,0 +1,50 @@
...
+                    {
+                        if (std::abs(center[2]) < 1e-12)
+                            cell->face(f)->set_boundary_id(1); // Front boundary --> h
+                        else if (std::abs(center[2] - 1.0) < 1e-12)
+                            cell->face(f)->set_boundary_id(1); // Back boundary ---> h
+                    }
+                }
+            }
+        }
+    }
```
michelebucelli, line 43: You can achieve the same result more efficiently setting the `colorize` argument of `hyper_cube` to `true`.

Also notice that in this loop you are not skipping over cells that are not locally owned (which may be present, due to internal details of the `parallel::distributed::Triangulation` class).

src/setup.cpp
```diff
@@ -0,0 +1,50 @@
+#include "setup.h"
+
+template<int dim>
+void setup_problem(
+    parallel::distributed::Triangulation<dim> &mesh,
+    FE_Q<dim>                                 &fe,
```
michelebucelli, line 6: Both the mesh and the finite element space are not meant to be modified by this function (and indeed they are not modified). As such, they should be passed as `const` arguments.

src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+    Utilities::MPI::MPI_InitFinalize mpi_init(argc, argv);
+    const unsigned int mpi_size = Utilities::MPI::n_mpi_processes(MPI_COMM_WORLD);
+    const unsigned int mpi_rank = Utilities::MPI::this_mpi_process(MPI_COMM_WORLD);
+    ConditionalOStream pcout(std::cout, mpi_rank == 0);
+
+    pcout << "Running with vectorization on " << VectorizedArray<double>::size() << " cells(" << Utilities::System::get_current_vectorization_level() << ")" << std::endl;
+    pcout << std::endl;
+
+    pcout << "Running matrix-free simulation with " << mpi_size << " processors" << std::endl;
+    auto start_time = std::chrono::high_resolution_clock::now();
```
michelebucelli, line 159: The `dealii::TimerOutput` class, and its member classes, offer a slightly more sophisticated interface to do this kind of thing.

src/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include "operator.h"
+#include "setup.h"
+#include "functions.h"
+#include "problem.h"
+
+using namespace dealii;
+using VectorType = LinearAlgebra::distributed::Vector<double>;
+
+template <int dim, int fe_degree, int ref_level>
+void run_simulation_matrix_free(ConditionalOStream &pcout){
```
michelebucelli, line 33: Is there any specific reason why the matrix-based solver is implemented using the object-oriented paradigm, whereas the matrix-free solver is not?

A good argument can be made for both (although the C++ language is to some extent designed around object-oriented programming). However, organizing the different parts of the code in a consistent manner is generally a good idea, and changes in paradigm should be well motivated.

In this particular instance, a good design could have been to have both the matrix-based and matrix-free solvers derive from one common base class, that implements the common parts (e.g. the management of the mesh). The derived classes would then implement assembly (where needed) and linear solver.

As a side note, consistency is especially important if you are comparing the performances of different solvers, as is the case here. I do not think that in this specific case the performance is impacted much, but it's better to err on the safe side.

include/operator.h
```diff
@@ -0,0 +1,57 @@
...
+                     const;
+
+private:
+
+    Table<2, VectorizedArray<double>> mu_coefficients;
+    //Table<3, VectorizedArray<double>> beta_coefficients;
+    Table<2, Tensor<1, dim, VectorizedArray<double>>> beta_coefficients;
+    Table<2, VectorizedArray<double>> gamma_coefficients;
+
+    const AffineConstraints<double> *constraints_ptr;
```
michelebucelli, line 54: References offer the same semantics as pointers, but with more safety. In this instance, a const reference would have worked better.

src/operator.cpp
```diff
@@ -0,0 +1,98 @@
...
+            phi.submit_value(gamma_coefficients(cell_batch, q) * value - beta_coefficients(cell_batch, q) * gradient, q);
+            phi.submit_gradient(mu_coefficients(cell_batch, q) * gradient, q);
+        }
+        phi.integrate(EvaluationFlags::values | EvaluationFlags::gradients);
+        phi.distribute_local_to_global(dst);
+    }
+}
+
+template<int dim, int fe_degree>
+void CustomOperator<dim, fe_degree>::compute_diagonal() {}
```
michelebucelli, line 95: This is potentially unsafe.

An user of your class (possibly you in the future) might think that this method is implemented and use it, get no error message, and yet get an incorrect behavior.

If you have to implement a method for interface compliance reasons, as in this case, but do not want to really implement it for whatever different reasons, it is a good idea to at least have the method throw an exception, so that, if someone tries to use it, they get told that the method is not implemented.

This is actually quite common, to the point that deal.II has an exception type, `ExcNotImplemented`, specifically for this kind of situation.

PR #18, opened 2025-02-06
NMPDE Wave Equation Solve - Luca Guffanti, Andrea Grassi, Nizare El Ziani

LucaGuffanti:
Hi, you find in the pull request the code for our project.

Thanks,
Luca, Andrea, Nizare


include/WaveParallel.hpp
```diff
@@ -0,0 +1,338 @@
...
+#include <iostream>
+#include <fstream>
+#include <memory>
+
+// ==================================================
+
+using namespace dealii;
+
+template <unsigned int dim>
+class WaveEquationParallel
```
michelebucelli, line 52: I am assuming that this class implements the Crank-Nicolson based solver, whereas the other one implements the solver based on Verlet integration.

If so, maybe the name of this class is not particularly explicative (also noting that there is no documentation comment that makes it obvious that this implements the Crank-Nicolson method).

include/WaveParallel.hpp
```diff
@@ -0,0 +1,338 @@
...
+    {
+    public:
+        BoundaryV(){}
+
+        virtual double value(const Point<dim>& /*p*/, const unsigned int /*component*/ = 0) const override
+        {
+            return 0.0;
+        }
+
+    };
```
michelebucelli, line 176: The problem does not involve any boundary conditions on the velocity auxiliary variable (since it inherits boundary conditions from the displacement variable). Are you sure that this class is necessary?

include/WaveParallel.hpp
```diff
@@ -0,0 +1,338 @@
...
+     * @tparam dim 
+     */
+    class ForcingTerm : public Function<dim>
+    {
+    public:
+        ForcingTerm(){}
+
+        virtual double value(const Point<dim>& p, const unsigned int /*component*/ = 0) const override
+        {
+            if (this->get_time() <= 0.5 && ((p[0]-0.5)*(p[0]-0.5) + (p[1]-0.5)*(p[1]-0.5)) <= 0.0025)
```
michelebucelli, line 190: As a general rule, I find it is a good idea to refrain from using "magic numbers" such as you do here (that is, hard-coding constants in function implementations).

A slightly better strategy would be to declare these numbers as `const` class members, initialized from the constructor. In the best-case scenario, you would read these numbers from some configuration file (for which, I encourage you to look into `dealii`'s `ParameterHandler` class).

include/WaveParallel.hpp
```diff
@@ -0,0 +1,338 @@
...
+     */
+    void setup(const std::string& mesh_filename);
+
+    /**
+     * @brief Constructs the triangulation, finite element space, DoF handler and linear algebra,
+     * utilising deal.ii mesh generation infrastructure.
+     * 
+     * @param times Number of times the mesh will be refined.
+     */
+    void setup(const unsigned int& times=1);
```
michelebucelli, line 68: As it turns out, these two methods do quite different things: the first reads a tetrahedral mesh from file, whereas the second builds a square mesh of hexahedra.

In this situation, I would refrain from relying on overloading. The most typical use case of overloading is when you have a function or method that does the same thing on different inputs. Here, you do different things with different inputs: it would probably be better to use different function names as well.

src/WaveParallel.cpp
```diff
@@ -0,0 +1,510 @@
...
+    laplace_matrix.compress(VectorOperation::add);
+
+    timer.exit_section("Assemble Mass/Laplace");
+}
+
+template <unsigned int dim>
+void WaveEquationParallel<dim>::run()
+{
+    timer.enter_section("Run");
+    timer.enter_subsection("Init");
```
michelebucelli, line 209: This can be made simpler and more C++-ish by using RAII through `TimerOutput::Scope`.

michelebucelli (commented):
Thanks for your submission!

As a general comment, I note that there's quite a bit of repeated code between the different implementations you propose.

While I understand that the serial implementation is sort of an intermediate step towards the parallel one (and thus would likely be discarded in a hypotetical production code), the Crank-Nicolson and Verlet solvers still have quite a bit of code in common that could have been externalized, e.g. to a common base class or to some smaller dedicated classes or functions.

Other than these, I have left a few small comments above.

PR #19, opened 2025-02-06
Consegna Progetto

leonardoevi:


src/Functions.hpp
```diff
@@ -0,0 +1,135 @@
+#ifndef FUNCTIONS_H
+#define FUNCTIONS_H
+
+#include <deal.II/base/function.h>
+#include "define.h"
+
+using namespace dealii;
+using namespace std;
+
+#if DIM == 1
```
michelebucelli, line 10: Preprocessor directives of this kind can and should usually be replaced with more modern (and less error-prone alternatives) such as `if constexpr` statements or template specializations.

src/define.h
```diff
@@ -0,0 +1,10 @@
+#ifndef DEFINE_H
+#define DEFINE_H
+
+
+#define DIM 2 // dimension of the problem to solve
+#define SKIPS 1 // Number of output time steps to skip
+#define LEAP_FROG false
+#define CONVERGENCE false // Performs convergence analysis
```
michelebucelli, line 8: Same as before on preprocessor directives.

To articulate a little better why that is, here you could e.g. write by mistake
```cpp
#define LEAP_FROG "false"
```
or any other non-boolean expression, leading to relatively hard to debug errors. Even worse, you might be tempted to write something like
```cpp
#define CONVERGENCE DIM == 2
```
which may (or may not) lead to incorrect expansions such as
```cpp
if (CONVERGENCE == true) -> if (DIM == 2 == true)
```

Conversely, more modern approaches to the same goal, like `if constexpr` statements, template specialization and other metaprogramming techniques, are usually more robust and overall better.

For a few of these flags, moreover, a run-time variable might have worked just fine.

src/WaveSolver.cpp
```diff
@@ -0,0 +1,541 @@
...
+  {
+    b = 0.0;
+
+    // b = M * v_k
+    mass_matrix.vmult(b, v_old_owned);
+
+    // tmp = delta_t * F_k
+    tmp = f_k;
+    tmp *= deltat;
+    b.add(tmp);
```
michelebucelli, line 276: These could have been achieved more efficiently through `b.add(deltat, f_k)`.

src/WaveSolver.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+    // tmp = A * u_k+1
+    stiffness_matrix.vmult(tmp, u_owned);
+    tmp *= (-deltat);
+    b.add(tmp);
```
michelebucelli, line 281: Same as above: `b.add(-deltat, tmp)`.

michelebucelli (commented):

PR #20, opened 2025-02-06
Fisher-Kolmogorov equation for neurodegenerative diseases

bralani:
Project members:

Balice Matteo
Doronzo Antonio Giuseppe
Sandretti Massimo Albino


src/main.cpp
```diff
@@ -0,0 +1,95 @@
...
+            } else if (w.length()==0 || w[0]=='#') {
+                // ignore empty lines and comments
+            } 
+            else {
+                std::cerr << "Invalid parameter: "<<w<<"\n";
+                exit(1);
+            }
+            
+        }
+    }
```
michelebucelli, line 66: You should consider looking into `dealii`'s `ParameterHandler` class for this sort of thing - it provides a much more streamlined and user-friendly interface.

src/DiffusionTensor.hpp
```diff
@@ -0,0 +1,116 @@
...
+public:
+    DiffusionTensor(double dext, double daxn) : dext(dext), daxn(daxn), identity(unit_symmetric_tensor<DIM>()) {}
+
+    Tensor<2, DIM> value(const Point<DIM> &p) const override {
+        return dext*identity + daxn*computeNxN(p);
+    }
+
+protected:
+    Tensor<2,DIM> computeNxN(const Point<DIM> &p) const {
+        Tensor<1,DIM> fiber_n = computeFiber(p);
```
michelebucelli, line 45:
```suggestion
        const Tensor<1,DIM> fiber_n = computeFiber(p);
```
might actually help a little bit with compiler optimization.

src/DiffusionTensor.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+using namespace dealii;
+
+template <unsigned int DIM>
+class DiffusionTensor : public TensorFunction<2, DIM> {
+public:
+    DiffusionTensor(double dext, double daxn) : dext(dext), daxn(daxn), identity(unit_symmetric_tensor<DIM>()) {}
+
+    Tensor<2, DIM> value(const Point<DIM> &p) const override {
+        return dext*identity + daxn*computeNxN(p);
```
michelebucelli, line 40: This function is probably called many times during assembly, and thus it might be worth it to optimize it a little bit.

In its current implementation, it requires the instantiation of at least three temporary `Tensor<2, DIM>` objects (one to store the result of `computeNxN`, one for each scalar-tensor multiplication), plus maybe one for the sum which however may be optimized through RVO.

Instantiating tensors is actually not so cheap (in relative terms, e.g. as opposed to instantiating 9 `double`s), and so you could consider replacing this line with the following:
```cpp
Tensor<2, DIM> result;
const Tensor<1, DIM> fiber_n = computeFiber(p);

for (unsigned int i = 0; i < DIM; ++i)
  {
    result[i][i] += dext;

    for (unsigned int j = 0; j < DIM; ++j)
      result[i][j] += daxn * fiber_n[i] * fiber_n[j];
  }
```

src/DiffusionTensor.hpp
```diff
@@ -0,0 +1,116 @@
...
+public:
+    RadialDiffusionTensor(double dext, double daxn, const Point<DIM> &radial_center) :
+        DiffusionTensor<DIM>(dext, daxn), radial_center(radial_center) {}
+
+
+protected:
+    Tensor<1,DIM> computeFiber(const Point<DIM> &p) const override {
+        Tensor<1,DIM> fiber_n;
+        for(int i = 0; i < DIM;i++) {
+            fiber_n[i] = (p[i]-radial_center[i])/(p.distance(radial_center) + 1.e-6);
```
michelebucelli, line 67: Same as above, an easy optimization would be to compute the denominator only once.

src/FisherKolmogorov.hpp
```diff
@@ -0,0 +1,252 @@
...
+
+  // Function for the forcing term.
+  class ForcingTerm : public Function<dim>
+  {
+  public:
+    virtual double
+    value(const Point<dim> & p,
+          const unsigned int /*component*/ = 0) const override
+    {
+      if(CONVERGENCE_TEST) {
```
michelebucelli, line 51: I would refrain from using define macros (except in very few occasions), since they are very error prone, for a number of reasons.

This flag could have been passed as a constructor argument, or you could have had two different function classes.

michelebucelli (commented):

PR #21, opened 2025-02-06
Consegna progetto Navier-Stokes Daneri-Vaglietti

giorgiodaneri:


src/test.cpp
```diff
@@ -0,0 +1,163 @@
...
+        {"solver", required_argument, 0, 's'},
+        {"tolerance", required_argument, 0, 't'},
+        {"preconditioner", required_argument, 0, 'p'},
+        {"help", no_argument, 0, 'h'},
+        {0, 0, 0, 0}
+    };
+
+    int opt;
+    // Modified getopt_long string to match the new format
+    while ((opt = getopt_long(argc, argv, "T:M:m:v:s:t:p:h", long_options, nullptr)) != -1) {
```
michelebucelli, line 52: I appreciate the fact that you provided a command line interface!

In the future, you might want to consider more C++-ish libraries than getopt, providing a better interface. For example, you might want to check out [clipp](https://github.com/muellan/clipp).

Additionally, it may be more convenient to provide some of the runtime parameters through a parameter file as supported by deal.II's ParameterHandler class.

src/testStationary.cpp
```diff
@@ -0,0 +1,146 @@
...
+              << "  -M, --read-mesh-from-file Provide mesh file path to load it instead or generating it inside the program\n"
+              << "  -m, --mesh-size X,Y       Set mesh size (two integers separated by a comma)\n"
+              << "  -v, --viscosity D         Set viscosity value (floating point value)\n"
+              << "  -s, --solver N            Select solver (valid values: 0: GMRES, 1: FGMRES, 2: Bicgstab)\n"
+              << "  -t, --tolerance D         Set tolerance (floating point value)\n"
+              << "  -p, --preconditioner N    Select preconditioner (valid values: 0: blockDiagonal, 1: blockTriangular, 2: aSIMPLE)\n"
+              << "  -h, --help                Display this help message\n";
+}
+
+int main(int argc, char *argv[]) {
```
michelebucelli, line 19: This file and test.cpp have a lot of repeated code, that could benefit from being externalized to a common place.

In medium to large projects, code repetition is a maintenance nightmare: if you find a bug in one of the copies of the code, you must remember to fix it also in the others, and you must be sure to do it exactly right. It is usually better to move the repeated parts to a common function (or class).

src/NSSolverStationary.hpp
```diff
@@ -0,0 +1,481 @@
...
+      if (re == 0.0)
+        u = 0.01;
+
+      if (u > U_m)
+        u = U_m;
+
+      return false;
+    }
+    const double U_m = 1.0;
+    const double H = 0.41;
```
michelebucelli, line 110: This is very minor: since these variables actually do not vary from one instance to the other of this class, you could consider making them `static constexpr`, which slightly facilitates compiler optimization and expresses more clearly the fact that these are "global" and not object-local.

src/NSSolverStationary.hpp
```diff
@@ -0,0 +1,481 @@
...
+    // Temporary vector.
+    mutable TrilinosWrappers::MPI::Vector tmp;
+  };
+
+  class PreconditionaSIMPLE {
+  public:
+    void initialize(const TrilinosWrappers::SparseMatrix &F_,
+                const TrilinosWrappers::SparseMatrix &B_,
+                const TrilinosWrappers::SparseMatrix &B_t_,
+                const TrilinosWrappers::MPI::BlockVector &vector_,
```
michelebucelli, line 245: As I understand it, this vector is only used to figure out the sizes and parallel distributions of the vectors that are needed by this class.

If that is correct, you can retrieve the same information through methods of the `SparseMatrix` class (e.g. `SparseMatrix::locally_owned_range_indices`, etc.).

src/NSSolverStationary.hpp
```diff
@@ -0,0 +1,481 @@
...
+            D_inv_vector[i] = 1.0 / D_vector[i];
+        }
+
+        // Compute Schur complement approximation: S = B * D^{-1} * B^T
+        S_matrix.clear();
+        
+        // Create S_matrix with appropriate sparsity pattern from B_t_
+        dealii::TrilinosWrappers::SparsityPattern sp(B_t_.m(), B_t_.m(), B_t_.n());
+        sp.compress();
+        S_matrix.reinit(sp);
```
michelebucelli, line 272: If I remember correctly, the `SparseMatrix::mmult` method also reinitializes the sparsity pattern of the destination matrix, so that these lines are probably superfluous. They are probably quite costly, however, and might partially explain why you find the aSIMPLE preconditioner to be slow in your tests (although I'm not entirely sure about this).

src/NSSolverStationary.hpp
```diff
@@ -0,0 +1,481 @@
...
+                     const unsigned int &degree_velocity_,
+                     const unsigned int &degree_pressure_,
+                     const unsigned int &mesh_size_x_,
+                     const unsigned int &mesh_size_y_,
+                     const unsigned int &solver_type_,
+                     const double &tolerance_,
+                     const unsigned int &preconditioner_type_,
+                     double nu_,
+                     bool read_mesh_from_file_)
+      : mpi_size(Utilities::MPI::n_mpi_processes(MPI_COMM_WORLD)), mpi_rank(Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)), pcout(std::cout, mpi_rank == 0), mesh_file_name(mesh_file_name_), mesh(MPI_COMM_WORLD), degree_velocity(degree_velocity_), degree_pressure(degree_pressure_), solver_type(solver_type_), tolerance(tolerance_), preconditioner_type(preconditioner_type_), mesh_size_x(mesh_size_x_), mesh_size_y(mesh_size_y_), nu(nu_), read_mesh_from_file(read_mesh_from_file_)
```
michelebucelli, line 349: Long lines such as this one are usually very hard to read, and thus may hide typos and errors. I suggest you setup your text editor of choice to apply automatic indentation, using packages such as `clang-format`.

src/NSSolverStationary.cpp
```diff
@@ -0,0 +1,933 @@
...
+        }
+        else
+        {
+          cell_data.material_id = 0;
+        }
+        cells.push_back(cell_data);
+      }
+
+      // Remove vertices that are not used in any cell.
+      GridTools::delete_unused_vertices(vertices, cells, subcell_data);
```
michelebucelli, line 66: There actually is a function named [`GridGenerator::channel_with_cylinder`](https://dealii.org/9.5.0/doxygen/deal.II/namespaceGridGenerator.html#a9d0b8a34fe5ca58425aa1fcf0d6132f1) that creates the mesh that you want, or at least something very similar to it.

src/NSSolverStationary.cpp
```diff
@@ -0,0 +1,933 @@
...
+
+        // assemble the residual associated with the nonlinear convective term
+        cell_rhs(i) -= scalar_product(nonlinear_term, fe_values[velocity].value(i, q)) * fe_values.JxW(q);
+
+        // b(v,p)
+        cell_rhs(i) += pressure_loc[q] *
+                       fe_values[velocity].divergence(i, q) *
+                       fe_values.JxW(q);
+
+        double velocity_divergence_loc = trace(velocity_gradient_loc[q]);
```
michelebucelli, line 489:
```suggestion
        const double velocity_divergence_loc = trace(velocity_gradient_loc[q]);
```
Always marking const the variables that are const may help with compiler optimization.

src/NSSolverStationary.cpp
```diff
@@ -0,0 +1,933 @@
...
+                                             boundary_values,
+                                             ComponentMask(
+                                                 {true, true, false}));
+
+    MatrixTools::apply_boundary_values(
+        boundary_values, jacobian_matrix, delta_owned, residual_vector, false);
+  }
+}
+
+int NSSolverStationary::solve_system() {
```
michelebucelli, line 579: This is the exact situation where polymorphism can be quite useful.

Instead of having 9 branches (3 types of preconditioners times 3 types of solvers), which may further increase if you add more, you could just declare a preconditioner through a pointer to the abstract base type (which is called `PreconditionBase` or something), instantiate it with the appropriate concrete type (3 if branches), and then forward it to the appropriate solver (3 other branches).

The overhead due to virtual dispatch is usually negligible, while the gain in code readability, modularity and thus maintainability is quite important.

src/NSSolverStationary.cpp
```diff
@@ -0,0 +1,933 @@
...
+  SolverControl solver_control(200000, tolerance);
+  // Choose the correct preconditioner
+  if (preconditioner_type == 0) {
+      PreconditionBlockDiagonal preconditioner;
+      preconditioner.initialize(jacobian_matrix.block(0, 0),
+                                    pressure_mass.block(1, 1));
+
+      if (solver_type == 0) {
```
michelebucelli, line 587: Additionally, you should consider using `enum class`es instead of "magic numbers" in this kind of situation. The underlying type (and thus the efficiency) is exactly the same, with a lot of added synctactic sugar.

src/NSSolverStationary.cpp
```diff
@@ -0,0 +1,933 @@
...
+  data_out.write_vtu_with_pvtu_record("./",
+                                      output_file_name,
+                                      0,
+                                      MPI_COMM_WORLD);
+
+  pcout << "Output written to " << output_file_name << std::endl;
+  pcout << "===============================================" << std::endl;
+}
+
+void NSSolverStationary::compute_lift_drag()
```
michelebucelli, line 802: [Functional programming](https://en.wikipedia.org/wiki/Functional_programming) is perhaps a bit dogmatic, but it does have some useful principles.

Among these is the idea that [pure functions](https://en.wikipedia.org/wiki/Pure_function) are often preferrable to functions that act by modifying the state. There's a few reasons for this, including better readability and overall safety (in terms of error proneness).

In this case, I think it would be better to have this function return its result (e.g. as an `std::pair`) rather than storing them in the corresponding class member. As a bonus, this would allow you to declare this method as const, which is conceptually more correct.

src/NSSolver.hpp
```diff
@@ -0,0 +1,551 @@
...
+#include <deal.II/numerics/matrix_tools.h>
+#include <deal.II/numerics/vector_tools.h>
+
+#include <fstream>
+#include <iostream>
+
+using namespace dealii;
+
+// Class representing the non-linear diffusion problem.
+class NSSolver
```
michelebucelli, line 46: Similarly to a comment I made before, there seems to be a substantial overlap between the steady and unsteady solvers. In production-level projects, you would probably want to move the repeated parts of the code to a common place, such as a common base class or something like that.

src/NSSolver.cpp
```diff
@@ -0,0 +1,1033 @@
...
+
+            // Pressure term in the momentum equation.
+            cell_matrix(i, j) -= fe_values[velocity].divergence(i, q) *
+                                 fe_values[pressure].value(j, q) *
+                                 fe_values.JxW(q);
+
+            // time dependent term (mass matrix)
+            cell_matrix(i, j) += (velocity_loc[q] - velocity_old_loc[q]) *
+                                 fe_values[velocity].value(i, q) / delta_t *
+                                 fe_values.JxW(q);
```
michelebucelli, line 399: This looks incorrect to me: shouldn't you have the value of the j-th shape function in place of `velocity_loc[q]`, and shouldn't the term related to the old velocity go to th e right-hand side only?

Notice how this contribution to the (i,j)-th entry of the local matrix does not depend on j: something is definitely wrong here.

src/NSSolverStationary.cpp
```diff
@@ -0,0 +1,933 @@
...
+    fe_values[pressure].get_function_values(solution, pressure_loc);
+
+    for (unsigned int q = 0; q < n_q; ++q)
+    {
+      for (unsigned int i = 0; i < dofs_per_cell; ++i)
+      {
+        for (unsigned int j = 0; j < dofs_per_cell; ++j)
+        {
+          if (global_first_iter || computing_stokes)
+          {
```
michelebucelli, line 384: Due to the similarity between Stokes and Navier-Stokes, this and the other branch have quite a few common terms (basically, all terms except for the transport term). It might be beneficial for the organization and clarity of the code to keep those common parts together.

michelebucelli (commented):

PR #22, opened 2025-02-06
Navier-Stokes  Solver. Bertasini-Cioci-Pagliochini.

aurorabertasini:
Dear Professor,
With this pull request, we submit our project.
Please notice that we used a separate branch to run numerical tests for all the different solvers implemented.


src/main.cpp
```diff
@@ -0,0 +1,115 @@
...
+        {
+            std::cin >> choice;
+            if (choice < 1 || choice > 6)
+            {
+                std::cout << "Invalid choice. Please enter a valid choice: ";
+            }
+        }
+    }
+
+    MPI_Bcast(&choice, 1, MPI_INT, 0, MPI_COMM_WORLD);
```
michelebucelli, line 56: `deal.II` offers a convenient, C++-ish interface to many MPI functions in the namespace `Utilities::MPI`.

src/main.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    auto start = std::chrono::high_resolution_clock::now();
+
+    switch (choice)
+    {
+    case 1:
```
michelebucelli, line 62: You should consider avoiding "magic numbers" such as these, and instead use more modern facilities of the C++ language such as `enum class`es.

src/main.cpp
```diff
@@ -0,0 +1,115 @@
...
+    {
+        if (mpi_rank == 0) std::cout << "Solving the Steady Navier-Stokesm Problem 2D" << std::endl;
```
michelebucelli, line 64: `deal.II` has a class named `ConditionalOStream` precisely for this purpose.

src/main.cpp
```diff
@@ -0,0 +1,115 @@
...
+            if (choice < 1 || choice > 6)
+            {
+                std::cout << "Invalid choice. Please enter a valid choice: ";
+            }
+        }
+    }
+
+    MPI_Bcast(&choice, 1, MPI_INT, 0, MPI_COMM_WORLD);
+
+    auto start = std::chrono::high_resolution_clock::now();
```
michelebucelli, line 58: You should consider using classes from `dealii::TimerOutput`, which can provide more fine-grained profiling.

include/ConfigReader.hpp
```diff
@@ -0,0 +1,49 @@
...
+    ConfigReader Class
+    ------------------
+    * Loads simulation config (mesh paths, polynomial degrees, simulation period, time step, Reynolds number) using regex and filesystem.
+    * Validates all required parameters and prompts user if any are missing.
+    * Provides public getters for easy access to these configuration values.
+    ------------------
+*/
+
+
+class ConfigReader
```
michelebucelli, line 19: There already exists a class within deal.II, named `ParameterHandler`, that allows to do this kind of thing. I suggest you look into it.

include/includes_file.hpp
```diff
@@ -0,0 +1,54 @@
+#ifndef INCLUDESFILE_HPP
```
michelebucelli, line 1: It is generally not a good idea to have a single file with all includes. Instead, in every file (both hpp and cpp) you should all and only the headers that are actually used within that file. This usually reduces compilation times and prunes the dependency graph that makefiles use to detremine what source files to recompile (which further reduces the compilation times).

src/MonolithicNavierStokes.cpp
```diff
@@ -0,0 +1,631 @@
...
+        {
+            for (unsigned int i = 0; i < dofs_per_cell; ++i)
+            {
+                for (unsigned int j = 0; j < dofs_per_cell; ++j)
+                {
+                    // Non-linear term
+                    // ------
+                    // N_ij(u*) = ∫ (u*·∇u)·φ_j dx
+                    // ------
+                    cell_lhs_matrix(i, j) += scalar_product(fe_values[velocity].gradient(j, q) * previous_velocity_values[q], fe_values[velocity].value(i, q)) * fe_values.JxW(q);
```
michelebucelli, line 288: Long lines such as this one are hard to read and to debug. I encourage you to install automatic formatting tools such as clang-format in your favorite text editor.

michelebucelli (commented):

PR #23, opened 2025-02-06
Project delivery

denisuntila:


src/NavierStokes.hpp
```diff
@@ -0,0 +1,355 @@
...
+
+
+using namespace dealii;
+
+
+template<size_t Dim>
+using FE_Type = FE_SimplexP<Dim>;
+
+template<size_t Dim>
+using Q_Type = QGaussSimplex<Dim>;
```
michelebucelli, line 49: Having type aliases such as these is a good idea! However, since these are conceptually bound to the `NavierStokes` class, it would probably be better to move these `using` declarations inside the class definition.

src/NavierStokes.hpp
```diff
@@ -0,0 +1,355 @@
...
+
+  class InletVelocity : public Function<dim>
+  {
+  public:
+    InletVelocity()
+      : Function<dim>(dim + 1)
+    {}
+
+
+    #ifndef NS_INPUT
```
michelebucelli, line 77: Using preprocessor macros for this conditional behavior should be your last resort, when all other options fail. Preprocessor macros may break C++ syntax and suffer from the same issues of global variables.

The alternatives include `if constexpr` statements, template specializations, or plain old `if` statements.

src/NavierStokes.hpp
```diff
@@ -0,0 +1,355 @@
...
+      else
+        return 0.0;
+    }
+  };
+
+  class PreconditionIdentity;
+
+  class PreconditionASIMPLE;
+  
+  class PreconditionAYosida;
```
michelebucelli, line 169: I don't think you actually need these forward declarations.

src/NavierStokes.hpp
```diff
@@ -0,0 +1,355 @@
...
+
+protected:
+};
+
+
+class NavierStokes::PreconditionASIMPLE
+{
+public:
+  // Application of the preconditioner: we just copy the input vector (src)
+  // into the output vector (dst).
```
michelebucelli, line 294: I think this comment is a left-over from copypasting.

Remember that "incorrect documentation is worse than no documentation".

src/NavierStokes.hpp
```diff
@@ -0,0 +1,355 @@
...
+
+};
+
+
+
+class NavierStokes::PreconditionAYosida
+{
+public:
+  // Application of the preconditioner: we just copy the input vector (src)
+  // into the output vector (dst).
```
michelebucelli, line 328: Same as above.

src/NavierStokes.cpp
```diff
@@ -0,0 +1,1050 @@
...
+  preconditioner.initialize(
+    system_matrix.block(0, 0),
+    system_matrix.block(1, 0),
+    system_matrix.block(0, 1),
+    deltat_lumped_mass_inv,
+    solution_owned
+  );
+  */
+  
+  auto end_time_prec = std::chrono::high_resolution_clock::now();
```
michelebucelli, line 375: `deal.II` offers the class `TimerOutput`, that provides profiling functionality with a much better interface.

src/NavierStokes.hpp
```diff
@@ -0,0 +1,355 @@
...
+  void export_data(const unsigned int &time_step);
+
+  void compute_ordered_dofs_indices();
+
+  void import_data(const unsigned int &time_step);
+
+  void post_process(const unsigned int &initial_time_step,
+    const unsigned int &final_time_step, const unsigned int &step);
+
+  void compute_forces(const double &time);
```
michelebucelli, line 218: For functions that compute stuff, unless there's strong reason to do otherwise, it is usually better to have the functions return their result, rather than have them store that result in a state variable (better in the sense of the code organization; this idea is borrowed from the principles of [functional programming](https://en.wikipedia.org/wiki/Functional_programming)).

michelebucelli (commented):
Thanks for your submission! I have left a few suggestions above.

PR #25, opened 2025-06-07
Project Submission

Valegrl:
Dear Professor,

With the following pull request we officially submit our project repository. Thank you.


1D/src/main_1D.cpp
```diff
@@ -0,0 +1,31 @@
...
+int main(int argc, char *argv[]) {
+  // Default values:
+  double d = 0.0001;
+  double alpha = 1.0;
+
+  // Check for command-line arguments.
+  if (argc > 1)
+    d = std::stod(argv[1]);
+  if (argc > 2)
+    alpha = std::stod(argv[2]);
```
michelebucelli, line 17: For your reference, there exist "smarter" (as in: more self-documenting, more C++-ish) ways of doing this. I and my group have been using [clipp](https://github.com/muellan/clipp) for a while and it works quite nicely.

1D/src/FisherKolmogorov1D.hpp
```diff
@@ -0,0 +1,198 @@
...
+    virtual Tensor<1, dim>
+    gradient(const Point<dim> &p,
+             const unsigned int /*component*/ = 0) const override {
+      Tensor<1, dim> result;
+
+      result[0] = -M_PI * std::sin(M_PI * p[0]) * std::exp(-get_time());
+
+      return result;
+    }
+  };
```
michelebucelli, line 85: You don't have an exact solution for your 1D test case, right? If so, this class is probably a leftover.

As a general rule, you should strive to keep your code clean, since leftovers such as this (or documentation errors, or whatever) can mislead people who might want to interpret your code in the future (possible your future selves as well).

(I know this is a small project, and it will most likely end here, but you should try to make these good practices a part of your instincts).

3D/src/ParameterReader.hpp
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * Declare all parameters that can be specified in the parameter file.
+   */
+  void declare_parameters();
+
+  /**
+   * Create the appropriate diffusion tensor based on parameters.
+   * @param params The simulation parameters
+   */
+  void create_diffusion_tensor(SimulationParameters &params);
```
michelebucelli, line 59: This is more of a "philosophical" comment, and may apply to this only in part, but still.

I find that some of the principles of [the functional programming paradigm](https://en.wikipedia.org/wiki/Functional_programming) are generally good to write well structured and maintainable code. When possible, I find it best to use pure functions (i.e functions without side-effects, semantically and synctactically). In this respect, in this case I would have preferred the function returned the diffusion tensor, rather than modify the state of its argument.

This is not fully agreed upon by everyone, and pure-functionalists - which by all means I am not - tend to be a bit dogmatic, so you may or may not agree with me. I just wanted to take the opportunity of mentioning this paradigm to you, so that you can take it into consideration for your future work.

3D/src/ParameterReader.hpp
```diff
@@ -0,0 +1,222 @@
...
+    prm.leave_subsection();
+
+    create_diffusion_tensor(params);
+
+  } catch (const std::exception &e) {
+    std::ostringstream oss;
+    oss << "Error with parameter file '" << parameter_file << "':\n"
+        << "Please check your parameter file format.";
+    throw std::runtime_error(oss.str());
+  }
```
michelebucelli, line 217: The `ParameterHandler::parse_input` method already performs a check on the consistency of the parameter file with the declared structure, and throws an exception with a meaningful error message if it does not. By meaningful here I mean that the message contains the problematic parameter or subsection, which makes it relatively easy to fix errors.

By manually catching the exception, you are losing this information here, and instead only stating which parameter file is causing issue (which however is an information that is likely already easy to access for the user).

3D/src/FisherKolmogorov3D.cpp
```diff
@@ -0,0 +1,295 @@
...
+      for (unsigned int i = 0; i < dofs_per_cell; ++i) {
+        for (unsigned int j = 0; j < dofs_per_cell; ++j) {
+          // Mass matrix.
+          cell_matrix(i, j) += fe_values.shape_value(i, q) *
+                               fe_values.shape_value(j, q) / deltat *
+                               fe_values.JxW(q);
+
+          // Non-linear stiffness matrix, first term.
+          cell_matrix(i, j) += d_loc * fe_values.shape_grad(i, q) *
+                               fe_values.shape_grad(j, q) * fe_values.JxW(q);
```
michelebucelli, line 160: This is probably not a problem here, but this statement is a potential source of inefficiency. That is because it needs to create a temporary `Tensor<1, dim>` to store the result of the first product, and the allocation and deallocation of this tensor might have a noticable impact on performance.

A faster version would look something like this:
```cpp
for (unsigned int c = 0; c < dim; ++c)
  for (unsigned int d = 0; d < dim; ++d)
    {
      cell_matrix(i, j) += d_loc[c][d]
                         * fe_values.shape_grad(i, q)[c]
                         * fe_values.shape_grad(j, q)[d]
                         * fe_values.JxW(q);
    }
```
Of course this comes at the price of readability, so you should weigh the efficiency gains against the performance improvements, but it's still something worth keeping in mind.

PR #26, opened 2025-06-07
Consegna Progetto Navier Stokes

Villanidav:
Davide Villani,  Emanuele Caruso, Sophia Peritz, Yosef Ben Chedly


src/Navier-Stokes/src/Convergence3D.cpp
```diff
@@ -0,0 +1,794 @@
...
+  const double tol = 1e-4 /**system_rhs.l2_norm()*/;
+  SolverControl solver_control(maxiter, tol, true);
+  // solver_control.enable_history_data();
+  SolverGMRES<TrilinosWrappers::MPI::BlockVector> solver(solver_control);
+  previous_solution = solution;
+  // Assemblying the preconditioner
+  {
+    dealii::Timer timerprec;
+    timerprec.restart();
+    dealii::Timer timersys;
```
michelebucelli, line 600: I find that classes `dealii::TimerOutput` and `dealii::TimerOutput::Scope` are much more convenient for profiling.

src/Navier-Stokes/src/Convergence3D.cpp
```diff
@@ -0,0 +1,794 @@
...
+            pcout << "Time taken to solve Navier Stokes problem: " << timersys.wall_time() << " seconds" << std::endl;
+            time_solve.push_back(timersys.wall_time());
+            
+            break;
+        }
+    
+        default:
+            throw std::runtime_error("Invalid preconditioner type");
+    }
+  }
```
michelebucelli, line 675: This code could be significantly simplified by having all preconditioner classes derive from the same base class with a standardized interface.

src/Navier-Stokes/src/main3D.cpp
```diff
@@ -0,0 +1,79 @@
...
+      
+    std::cin >> test_case;  // Read user input into test_case
+    if (test_case <= 1 || test_case > 3) 
+          {
+              std::cerr << "Invalid test case number. Using default (2)." << std::endl;
+              test_case = 2;
+          }    
+  }*/
+
+  MPI_Bcast(&test_case, 1, MPI_INT, 0, MPI_COMM_WORLD);
```
michelebucelli, line 28: You should look into the namespace `dealii::Utilities::MPI`, which offers a few functions that improve the interface of MPI calls.

Also, I imagine that this broadcast call is a leftover from when the code commented above was not commented out. The task in the commented code would be achieved much better through command line arguments (possibly through a high-level library such as [clipp](https://github.com/muellan/clipp)) or through parameter files (see the class `dealii::ParameterHandler`).

michelebucelli (commented):
As a general comment, you have a good deal of repeated code between the different variants of your solver (2D, 3D, conergence studies).

In a situation like this, you should consider moving the repeated code to a common part (e.g. a general-purpose, dimension-independent Navier-Stokes solver), to avoid the overlap.

Such an overlap quickly becomes a nightmare to maintain, and you should try to avoid it as much as possible. I know this project is limited in scope, but you should take projects such as this as opportunities to ingrain good development practices into your programming habits.

PR #27, opened 2025-06-07
Project presentation - June

MatteoArrigo:
Pull request per la presentazione del progetto di Giugno
Gruppo: Arrigo, Noventa, Severino
Progetto: n. 5, Fisher-Kolmogorov equation for neurodegenerative diseases

### DIFF
--- a/src/1d/FisherKolmogorov1d.cpp
+++ b/src/1d/FisherKolmogorov1d.cpp
@@ -1,0 +1,276 @@
+#include "FisherKolmogorov1d.hpp"
+
+void FisherKolmogorov1D::setup() {
+    namespace fs = std::filesystem;
+    try {
+        if (fs::exists(output_directory)) {
+            fs::remove_all(output_directory); // Deletes the entire directory
+        }
+        fs::create_directories(output_directory); // Recreates the directory
+    } catch (const std::exception &e) {
+        std::cerr << "Error in directory creation: " << e.what() << std::endl;
+    }
+
+    // Create the mesh.
+    {
+        std::cout << "Initializing the mesh" << std::endl;
+        GridGenerator::subdivided_hyper_cube(mesh, N, -1.0, 1.0, true);
+        std::cout << "  Number of elements = " << mesh.n_active_cells()
+                  << std::endl;
+
+        // Write the mesh to file.
+        const std::string mesh_file_name =
+            output_directory + "/mesh-" + std::to_string(N) + ".vtk";
+        GridOut grid_out;
+        std::ofstream grid_out_file(mesh_file_name);
+        grid_out.write_vtk(mesh, grid_out_file);
+        std::cout << "  Mesh saved to " << mesh_file_name << std::endl;
+    }
+
+    std::cout << "-----------------------------------------------" << std::endl;
+
+    // Initialize the finite element space.
+    {
+        std::cout << "Initializing the finite element space" << std::endl;
+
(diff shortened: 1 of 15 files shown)


src/1d/FisherKolmogorov1d.cpp
```diff
@@ -0,0 +1,276 @@
+#include "FisherKolmogorov1d.hpp"
+
+void FisherKolmogorov1D::setup() {
+    namespace fs = std::filesystem;
+    try {
+        if (fs::exists(output_directory)) {
+            fs::remove_all(output_directory); // Deletes the entire directory
```
michelebucelli, line 7: While I understand that this is convenient, it is generally a bad idea to have a program quietly delete the whole content of a directory.

src/3d/DiffusionType.hpp
```diff
@@ -0,0 +1,32 @@
...
+ * The DiffusionType enum is used to distinguish between different anisotropic diffusion behaviors:
+ * - ISOTROPIC: Uniform diffusion in all directions.
+ * - CIRCUMFERENTIAL: Diffusion primarily in the circumferential direction.
+ * - RADIAL: Diffusion primarily in the radial direction.
+ * 
+ * Note: We did not implement the axon-based anisotropic diffusion model, due to lack of real-world data.
+ */
+
+
+enum DiffusionType {
```
michelebucelli, line 16: You should consider using `enum class`es in place of `enum`s, since they are more type-safe.

src/3d/FisherKolmogorov3d.cpp
```diff
@@ -0,0 +1,292 @@
...
+                    // Mass matrix.
+                    cell_matrix(i, j) += fe_values.shape_value(i, q) *
+                                         fe_values.shape_value(j, q) / deltat *
+                                         fe_values.JxW(q);
+
+                    // Stiffness matrix, Diffusion term (linear)
+                    cell_matrix(i, j) +=
+                        fe_values.shape_grad(i, q) *
+                        d_loc *
+                        fe_values.shape_grad(j, q) *
```
michelebucelli, line 154: This is probably not a problem here, but this statement is a potential source of inefficiency. That is because it needs to create a temporary `Tensor<1, dim>` to store the result of the first product, and the allocation and deallocation of this tensor might have a noticable impact on performance.

A faster version would look something like this:
```cpp
for (unsigned int c = 0; c < dim; ++c)
  for (unsigned int d = 0; d < dim; ++d)
    {
      cell_matrix(i, j) += d_loc[c][d]
                         * fe_values.shape_grad(i, q)[c]
                         * fe_values.shape_grad(j, q)[d]
                         * fe_values.JxW(q);
    }
```
Of course this comes at the price of readability, so you should weigh the efficiency gains against the performance improvements, but it's still something worth keeping in mind.

src/3d/SimulationParams.hpp
```diff
@@ -0,0 +1,86 @@
...
+ * @note
+ * - The CSV file must not contain a header row.
+ * - Each line should be of the form: key,value
+ * - Whitespace around keys and values is ignored.
+ * - All required keys must be present in the CSV file.
+ *
+ */
+
+
+struct SimulationParams {
```
michelebucelli, line 62: Have you looked into `dealii::ParameterHandler` for the same purpose? It offers a nicely structured way of reading parameters from file, checking the consistency of the parameter file against a fixed schema to make sure everything is correct.

PR #28, opened 2025-06-07
Submission

edo01:
Fisher-Kolmogorov equation for neurodegenerative diseases. Group composed by: Allanda Marco, Carrà Edoardo, Ferrario Daniele, Gentile Lorenzo.


src/common/NDConfig.cpp
```diff
@@ -0,0 +1,31 @@
...
+        case 's': this->seeding_region_type = SeedingRegionType(std::atoi(optarg)); break;
+        case 'f': this->fiber_field_type = FiberFieldType(std::atoi(optarg)); break;
+        case 'h':
+            std::cerr << "Usage: " << argv[0] << " [options]\n";
+            exit(EXIT_SUCCESS);
+        default:
+            std::cerr << "Invalid option\n";
+            exit(EXIT_FAILURE);
+      }
+    }
```
michelebucelli, line 30: For your reference, there are much cleaner ways of doing this. You can look into [clipp](https://github.com/muellan/clipp), or use `dealii::ParameterHandler` to read configuration from file (or at the very least use getpot in place of getopt).

include/WhiteGrayPartition.hpp
```diff
@@ -0,0 +1,82 @@
...
+            {
+                if (current_cell_center.distance(boundary_center) < distance_threshold)
+                {
+                    cell->set_material_id(1); // Set the material ID to 0 for gray matter
+                    is_gray = true;
+                    break;
+                }
+            }
+            if (!is_gray) cell->set_material_id(0); 
+        }
```
michelebucelli, line 52: This is very inefficient, due to the double nested loop. Its complexity scales more or less like $O(n^2)$, with $n$ being the number of vertices in the mesh. You should look into R-Trees for a much faster way of doing this kind of thing.

PR #29, opened 2025-06-09
project delivery

francescapieruz16:


src/WaveFunction.cpp
```diff
@@ -0,0 +1,256 @@
...
+  using namespace dealii;
+
+  template <int dim>
+  WaveEquation<dim>::WaveEquation()
+    : fe(1)                                       // finite element of degree 1
+    , dof_handler(triangulation)                 // initialize the degree of freedom handler with the mesh
+    , time_step(1. / 64)                         // time step size
+    , time(time_step)                            // current simulation time (starts at one time step)
+    , timestep_number(1)                         // counter for time steps
+    , theta(0.5)                                 // theta method parameter (0.5 = Crank-Nicolson)
```
michelebucelli, line 16: There's a number of hard-coded constants here (and elsewhere) which would make it somewhat cumbersome to test different configurations of your solver (e.g. different discretization sizes, polynomial degrees, ...).

A better design choice would be to pass these from outside, either as constructor argument or through some external parameter file (cf. `dealii::ParameterHandler`).

src/WaveFunction.cpp
```diff
@@ -0,0 +1,256 @@
...
+    VectorTools::project(dof_handler, constraints, QGauss<dim>(fe.degree + 1),
+                         InitialValuesU<dim>(), old_solution_u);
+    VectorTools::project(dof_handler, constraints, QGauss<dim>(fe.degree + 1),
+                         InitialValuesV<dim>(), old_solution_v);
+
+    Vector<double> tmp(solution_u.size());
+    Vector<double> forcing_terms(solution_u.size());
+
+    // time-stepping loop from t = 0 to t = 5
+    for (; time <= 5; time += time_step, ++timestep_number)
```
michelebucelli, line 152: This literal `5` is another "magic constant": this is a very bad practice, you should at the very least declare a `const` or `constexpr` variable somewhere.

PR #30, opened 2025-07-07
A pull request for the nmpde project

Amber12355:
This pull request includes all prior work migrated from our earlier repository into this new fork.


.DS_Store
michelebucelli, line 1: Minor but useful hint: this file (and other files with the same name) is a MacOS-specific file used to store folder properties or similar. It should not be pushed to a git repository.

The simplest way to avoid accidentally pushing it is to [setup a `.gitignore` file](https://docs.github.com/en/get-started/git-basics/ignoring-files).

FishKolm/mesh/brain-h3.0.msh
michelebucelli, line 1: As indicated in [these instructions](https://github.com/michelebucelli/nmpde-projects/blob/main/README.md), you should not have uploaded the mesh files to the git repository.

The reason is that git is not suited to managing large binary (or binary-like) files, due to its internals. You should look into git large file storage (LFS) for your future work.

FishKolm/src/FishKolm.hpp
```diff
@@ -0,0 +1,324 @@
...
+      result[0] = 0;
+      result[1] = 0;
+      if (dim == 3)
+      {
+        result[2] = 0;
+      }
+
+      return result;
+    }
+  };
```
michelebucelli, line 187: I'm guessing you don't really have an exact solution (and that solution would not be zero), so this class is actually not useful.

As a general rule, you should remove this sort of leftovers, as they can be misleading.

FishKolm/src/FishKolm.cpp
```diff
@@ -0,0 +1,388 @@
...
+      const auto &x_q = fe_values.quadrature_point(q);
+      const double JxW = fe_values.JxW(q);
+
+      const double alpha = alpha_coefficient.value(x_q);
+      const Tensor<2, dim> D_matrix = D.matrix_value(x_q);
+
+      for (unsigned int i = 0; i < dofs_per_cell; ++i)
+      {
+        const double phi_i = fe_values.shape_value(i, q);
+        const Tensor<1, dim> grad_phi_i = fe_values.shape_grad(i, q);
```
michelebucelli, line 141: Somewhat counterintuitively, this is less efficient than directly calling `fe_values.shape_value(i, q)` and `fe_values.shape_grad(i, q)` directly in the inner loop (the loop over `j`).

The reason is that `FEValues::shape_value` and `shape_grad` actually return a reference to the shape function evaluation (or its gradient) that is stored internally to the `FEValues` instance. No computations are done when you call `shape_value` or `shape_grad`, everything is precomputed when calling `FEValues::reinit`.

Therefore, these two lines are inefficient because they allocate new memory and create copies of data that could already be accessed.

FishKolm/src/FishKolm.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+        for (unsigned int j = 0; j < dofs_per_cell; ++j)
+        {
+          const double phi_j = fe_values.shape_value(j, q);
+          const Tensor<1, dim> grad_phi_j = fe_values.shape_grad(j, q);
```
michelebucelli, line 146: Same as above.

FishKolm/src/FishKolm.cpp
```diff
@@ -0,0 +1,388 @@
...
+{
+  SolverControl solver_control(5000, 1e-6 * system_rhs.l2_norm());
+  SolverCG<TrilinosWrappers::MPI::Vector> solver(solver_control);
+  TrilinosWrappers::PreconditionAMG preconditioner;
+  TrilinosWrappers::PreconditionAMG::AdditionalData amg_data;
+  preconditioner.initialize(lhs_matrix, amg_data);
+
+  solver.solve(lhs_matrix, delta_owned, system_rhs, preconditioner);
+  pcout << "  " << solver_control.last_step() << " CG iterations" << std::endl;
+}
```
michelebucelli, line 200: This is quite an inelegant way of achieving this. Indeed, there's a bit of repeated logic between the two methods, and even more importantly the methods' names do not clearly indicate what the difference is between the two (how was I to know that `solve_linear_system_2` uses AMG instead of SSOR?).

A more appropriate way to go would be to
- use class member variables to store the requested solver control parameters;
- use polymorphism to allow for different preconditioners, together with a variable (e.g. a string, or even better an `enum class`) that allows selecting between AMG and SSOR.

FishKolm/src/FishKolm.cpp
```diff
@@ -0,0 +1,388 @@
...
+    ++n_iter;
+  }
+
+  if (n_iter == n_max_iters && residual_norm > residual_tolerance)
+  {
+    pcout << "  Warning: Newton solver did not converge after "
+          << n_max_iters << " iterations. Final residual: "
+          << residual_norm << std::endl;
+  }
+}
```
michelebucelli, line 275: Similar to the above: what is the difference between these two methods? Does it really make sense to keep both, or is there one that is strictly better than the other, so that the latter can be discarded altogether?

FishKolm/src/FishKolm.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+    solution_old = solution;
+
+    pcout << "n = " << std::setw(3) << time_step << ", t = " << std::setw(5)
+          << time << ":" << std::endl;
+
+    solve_newton();
+    output(time_step);
+  }
+}
```
michelebucelli, line 370: Somewhat related to the previous: here the difference between `solve_opt` and `solve` is that the former uses time adaptivity. Then, wouldn't `solve_adaptive` be a more self-documenting name?

FishKolm/src/FishKolm.cpp
```diff
@@ -0,0 +1,388 @@
...
+    pcout << "  Warning: Newton solver did not converge after "
+          << n_max_iters << " iterations. Final residual: "
+          << residual_norm << std::endl;
+  }
+}
+void 
+FISHKOLM::solve_opt()
+{
+  Timer timer;
+  timer.start();
```
michelebucelli, line 280: I find that using `dealii::TimerOutput::Scope` is more flexible for this kind of profiling.

PR #37, opened 2025-12-22
Wave Equation 

Oleg-Nedina:
This is my first time submitting a Pull Request, so please accept my apologies if I accidentally created extra forks. The code follows a structure very similar to the one we saw in the laboratory. I look forward to your feedback.


include/input_parser.hpp
```diff
@@ -0,0 +1,39 @@
...
+        unsigned int get_scenario_id() const;
+        
+        std::string get_solver_type() const; 
+
+        double get_beta() const;
+        double get_gamma() const;
+
+    private:
+        // 'mutable' permette ai metodi const di modificare lo stato interno
+        mutable ParameterHandler prm;
```
michelebucelli, line 32: I assume you declared this as `mutable` so that getter methods of this class could access the parameter values stored inside the handler.

It would probably be better to read the parameter values already inside the `parse_parameters` method and storing those values as class members, so that the getters would only return the value of those members, and `prm` would not need to be declared as `mutable`. This would also be somewhat more efficient, since every time you read a parameter from `ParameterHandler` there's a string-to-something conversion involved.

include/input_parser.hpp
```diff
@@ -0,0 +1,39 @@
...
+        // Funzione principale di parsing
+        void parse_parameters(const std::string &filename);
+
+        // Getters
+        double get_time_step() const;
+        double get_final_time() const;
+        unsigned int get_output_frequency() const;
+        unsigned int get_scenario_id() const;
+        
+        std::string get_solver_type() const;
```
michelebucelli, line 25: This kind of information is better represented through an `enum class`, instead of an `std::string`.

main.cpp
```diff
@@ -0,0 +1,51 @@
...
+        Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);
+
+        // Controllo argomenti:
+        // Se l'utente passa un file, lo usa. Altrimenti cerca "parameters.prm" di default.
+        std::string parameter_file;
+        if (argc > 1) {
+            parameter_file = argv[1];
+        } else {
+            parameter_file = "parameters.prm";
+        }
```
michelebucelli, line 21: For the future, I suggest using more structured tools for parsing command line arguments. A nice example is [clipp](https://github.com/muellan/clipp), but there are quite many options.

include/test_cases.hpp
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * @brief Exact analytical solution u_ex(x,t).
+     */
+    template <int dim>
+    class ExactSolution_Case1 : public Function<dim> {
+    public:
+        ExactSolution_Case1() : Function<dim>(1) {} 
+        
+        virtual double value(const Point<dim> &p, const unsigned int /*component*/ = 0) const override {
+            double time = this->get_time();
```
michelebucelli, line 34:
```suggestion
            const double time = this->get_time();
```
If something is not supposed to change, you should always declare it as `const`. This helps a lot with compiler optimization, which in many cases leads to measurably better performance. This is probably not one of those cases, but you should get in the habit of using `const` appropriately as part of your "coding hygiene".

include/test_cases.hpp
```diff
@@ -0,0 +1,251 @@
...
+using namespace dealii;
+
+namespace WaveEquationProject {
+
+
+    // =========================================================================
+    // SCENARIO 1: MANUFACTURED SOLUTION (Dim-Independent)
+    // =========================================================================
+    // Description:
+    // Uses a smooth analytical function to verify the order of convergence.
```
michelebucelli, line 17: Minor: I notice that some of the comments are in Italian, others are in English. You should try to be as consistent as possible, especially for collaborative work.

wave_equation/wave_equation.cpp
```diff
@@ -0,0 +1,1042 @@
...
+    this->method_type = EXPLICIT;
+  }
+  this->beta = parser.get_beta();
+  this->gamma = parser.get_gamma();
+  pcout << "------------------------------------------" << std::endl;
+  pcout << "Parameters loaded from: " << param_file << std::endl;
+  pcout << "Method: " << solver_str << " | dt: " << time_step
+        << " | T: " << final_time << " | Scenario: " << scenario_id
+        << std::endl;
+  pcout << "------------------------------------------" << std::endl;
```
michelebucelli, line 72: Logically, this should be part of the `InputParser` class.

Currently the parameters are managed in part by `InputParser` and in part by `WaveEquation`: this means that there is overlap between the responsibilities of the two classes, which violates the principle of separation of concerns.

wave_equation/wave_equation.cpp
```diff
@@ -0,0 +1,1042 @@
...
+
+  // Create a unit square [0,1]^d
+  GridGenerator::hyper_cube(triangulation, 0, 1);
+
+  int tmp_int_ref;
+  if (dim == 2) {
+    tmp_int_ref = 6;
+  } else {
+    tmp_int_ref = 4;
+  }
```
michelebucelli, line 158: This information should probably be read from the parameter file.

PR #40, opened 2025-12-23
Project 5. Space and Time Adaptivity

GeckyMerlo:
The project was carried out by the following group members:
	•	Bianca Maria Marchettini – 11098150
	•	Aurora Perlini – 10788118
	•	Domenico De Giorgio – 10854350
	•	Giacomo Merlo – 10848938


src/main.cc
```diff
@@ -0,0 +1,245 @@
...
+      if (cells_per_direction <= 0)
+      {
+        std::cerr << "Invalid number of cells. It must be a positive integer.\n";
+        return 1;
+      }
+      std::cout << "Generating a mesh with " << cells_per_direction << " cells per direction.\n";
+    }
+
+    // --- Equation Physical Parameters ---
+    const double T_end       = ask_double_default("Enter T (final time)", 0.5);
```
michelebucelli: I appreciate that you allow the user to provide these without having to recompile the source. 

However, this prompt-based strategy does not translate well to large scale computing situations, where you would typically submit a script to a scheduler, which queues it to other jobs. In those situations, it is usually better to provide the settings through a parameter file (which also has a few other quality-of-life bonuses: imagine that the user makes a mistake at the 15th prompt, then they have to restart from scratch).

I suggest you look into `dealii::ParameterHandler` for this kind of thing.

src/heat_equation.cc
```diff
@@ -0,0 +1,585 @@
...
+    time_step_min       = time_step_min_in;
+
+    end_time = end_time_in;
+    time_step = initial_dt_in;
+
+    initial_global_refinement = initial_global_refinement_in;
+    n_adaptive_pre_refinement_steps = n_adaptive_pre_refinement_steps_in;
+    refine_every_n_steps = refine_every_n_steps_in;
+
+    write_vtk = write_vtk_in;
```
michelebucelli: Most of these these should have been included in the initializer list.

src/heat_equation.cc
```diff
@@ -0,0 +1,585 @@
...
+    system_rhs.reinit(dof_handler.n_dofs());
+  }
+
+  template <int dim>
+  void HeatEquation<dim>::solve_time_step()
+  {
+    const double rhs_norm = system_rhs.l2_norm();
+    const double tol = std::max(1e-14, 1e-8 * rhs_norm);
+
+    SolverControl            solver_control(1000, tol);
```
michelebucelli: There's a class named `ReductionControl` that is a better way of doing this.

src/heat_equation.cc
```diff
@@ -0,0 +1,585 @@
...
+    SolverCG<Vector<double>> cg(solver_control);
+
+    preconditioner.initialize(system_matrix, 1.0);
```
michelebucelli: The matrix only changes between timesteps if the timestep changes. Therefore, you might want to only initialize the preconditioner when the matrix changes, to save some computations.

PR #42, opened 2025-12-23
Project 2 evaluation request

supertramp277:
Dear prof:
We have finished the project 2 about wave equation with a lot different test containing both time and spatial convergence, dissipation and dispersion study. We also submit the plots and result files. The folder ./result/dissipation/energy and ./result/dispersion/energy contain a lot files which are a little boring. We add them just for completenss. Instead, the summary files of them are much more meaningful. Rest details about compiling and running are already in the readme. Here are group information:
- Yanlong Wang
- Zhaochen Qiao
- Luca Tramacere
We are looking forward to your feedback and improve it further. Thanks a lot!


include/IOUtils.hpp
```diff
@@ -0,0 +1,94 @@
...
+    {
+        if (dir.empty() || dir == ".")
+            return;
+
+        int ok = 1;
+
+        if (mpi_rank == 0)
+        {
+            errno = 0;
+            const int rc = ::mkdir(dir.c_str(), 0755);
```
michelebucelli, line 32: Have you considered using the C++-17 standard library header [`filesystem`](https://en.cppreference.com/w/cpp/filesystem.html) instead?

PR #44, opened 2025-12-23
 NMPDE Project delivery - Wave Equation

UobinoPino:


michelebucelli:
Please update the description or the title with the group members. I can try and guess who you are from the nicknames, but I'd rather not to.

src/Newmark/main.cpp
```diff
@@ -0,0 +1,98 @@
...
+                   /* degree = */ 1,
+                   /* T = */ 2.0,
+                   /* beta = */ 0.25, //0.25
+                   /* gamma = */ 0.5,
+                   /* delta_t = */ 0.01,
+                   /* test_case = */ test_case);
+
+      // Synchronize before timing
+      MPI_Barrier(MPI_COMM_WORLD);
+      auto start = std::chrono::high_resolution_clock::now();
```
michelebucelli: Have you considered using `dealii::TimerOutput` for this?

src/Newmark/Wave.hpp
```diff
@@ -0,0 +1,402 @@
+#ifndef WAVE_HPP
+#define WAVE_HPP
```
michelebucelli: This headerguard is the same as the one of `src/Theta_Method/serial/Wave.hpp`. As a general rule, headerguards should be unique to each file, because their purpose is to avoid double inclusion of that file.

Having the same headerguard in two files means that you might not be able to include them both, although if you tried to do it you wouldn't get any error message, leading to all sorts of hard-to-diagnose incorrect behavior.

src/Newmark/main.cpp
```diff
@@ -0,0 +1,98 @@
...
+      std::unique_ptr<Function<dim>> exact_solution;
+
+      if (test_case == Wave::EX1)
+        {
+          exact_solution = std::make_unique<Wave::ExactSolutionEX1>();
+          if (mpi_rank == 0)
+            {
+              std::cout << "Running EX1: u(x,y,t) = sin(π(x+1)/2) * sin(π(y+1)/2) * cos(t)" << std::endl;
+              std::cout << "             f = (π²/2 - 1) * φ(x,y) * cos(t)" << std::endl;
+            }
```
michelebucelli: Using a `ConditionalOStream` makes this kind of thing much cleaner.

michelebucelli:
A couple general observations:

1. there is considerable overlap between the three solvers (serial $\theta$-method, parallel $\theta$-method, Newmark method), which could have been managed slightly better by centralizing the common parts to a common, base class (or equivalent);
2. it might have been a good idea to devise a method for the user to input problem settings without having to modify the source code. Some other groups chose to use command line arguments (for which I recommend using a high-level tool such as [clipp](https://github.com/muellan/clipp)), others used `dealii::ParameterHandler`, some used a combination of both. Regardless, it is generally a better idea to have a way of configuring the solver without the need to edit the source files and recompile.

src/Newmark/main.cpp
UobinoPino: No, we didn't. We now use it

src/Newmark/Wave.hpp
UobinoPino: Thanks!

src/Newmark/main.cpp
UobinoPino: Thank you for the tip, we have now updated the code

PR #24, opened 2025-06-06
merge branch master with main 

francescapieruz16:

PR #45, opened 2025-12-23
[Project 2 - Solving the wave equation] Abril Cano Castro, Alberto Ondei, Andrea Valentini, Giuseppe Starnini

AndreVale69:
Group members (name surname - ID):

- Abril Cano Castro - 241744
- Alberto Ondei - 277739
- Andrea Valentini - 260236
- Giuseppe Starnini - 252540


AndreVale69:
Also @abrilcano and @Sterning171177 make commits in other branches that are manually merged (not in Git).

michelebucelli:
Please update the description or the title with the names of the group members. I can guess who you are from the GitHub nicknames, but I'd rather not to. Remember that this delivery is part of an exam, and communicate with the appropriate formality.

AndreVale69:
> Please update the description or the title with the names of the group members. I can guess who you are from the GitHub nicknames, but I'd rather not to. Remember that this delivery is part of an exam, and communicate with the appropriate formality.

Thank you for the clarification, Professor.

We updated the PR title and description to include the full names and student IDs of the group members. We apologize for the oversight and thank you for the reminder about the appropriate level of formality.

include/enum/boundary_type.hpp
```diff
@@ -0,0 +1,63 @@
...
+        case BoundaryType::Zero:
+            return "zero";
+        case BoundaryType::MMS:
+            return "mms";
+        case BoundaryType::Expr:
+            return "expr";
+        default:
+            return "unknown";
+    }
+}
```
michelebucelli, line 30: You might want to check out [magic_enum](https://github.com/Neargye/magic_enum), I've never used it personally but it's supposed to automate this kind of thing. Also, it is bundled with `deal.II` (although I'm not sure in which version this started - I think 9.6 or 9.7).

Also, it might be a better idea to raise an exception upon encountering an unknown value, rather than returning a placeholder string.

include/functions/mms_functions.hpp
```diff
@@ -0,0 +1,108 @@
...
+     * @details
+     * The expression is of the form:
+     * sin(pi*x)*sin(pi*y)*cos(omega*t)
+     */
+    std::string get_expression() const {
+        return "sin(pi*x)*sin(pi*y)*cos(" + std::to_string(omega) + "*t)";
+    }
+
+private:
+    double omega;
```
michelebucelli: This should be declared as `const`.

include/parameters.hpp
```diff
@@ -0,0 +1,824 @@
...
+        declare();
+        parse(filename);
+    }
+
+    /**
+     * @brief Initialize the problem functions (mu, forcing term, initial conditions).
+     * @tparam dim Spatial dimension.
+     * @param mu FunctionParser for the mu coefficient.
+     * @param boundary_g FunctionParser for the boundary condition g.
+     * @param boundary_v FunctionParser for the boundary condition for the velocity.
```
michelebucelli: These are not FunctionParser but just functions.

src/main.cpp
```diff
@@ -0,0 +1,26 @@
...
+    const std::filesystem::path main_file_path = std::filesystem::canonical(argv[0]).parent_path();
+    wave_prm = (main_file_path / wave_prm).string();
+    if (argc > 1) {
+        wave_prm = argv[1];
+    } else {
+        std::cout << "No parameter file provided. Run with mpirun -np <nprocs> ./wave "
+                     "<parameter_file>"
+                  << std::endl;
+        std::cout << "Using default parameter file: " << wave_prm << std::endl;
+    }
```
michelebucelli: This is a bit of a counterintuitive implementation. The following would be equivalent but somewhat more natural:
```cpp
    std::string wave_prm;

    if (argc > 1) {
        wave_prm = argv[1];
    } else {
        const std::filesystem::path main_file_path = std::filesystem::canonical(argv[0]).parent_path();
        wave_prm = (main_file_path / wave_prm).string();
    }

```

src/utils/mesh_generator.cpp
```diff
@@ -0,0 +1,95 @@
+#include "utils/mesh_generator.hpp"
+
+#include <cstdlib>
+#include <cstring>
+#include <fstream>
+#include <stdexcept>
+
+namespace mesh_generator {
```
michelebucelli, line 8: I'm not sure if that does what you want, but `deal.II` does have a gmsh interface (and gmsh has a C++ API that most likely can do what you want).

include/utils/progress_bar.hpp
michelebucelli, line 1: This is quite interesting!

However, you should be aware that it most likely breaks in a cluster setting (although of course you would have to test it). When running large jobs on a cluster, one would usually submit a job script to a scheduler, and the script would be executed at some later time. To view the output, the script usually redirects standard output and standard error to file, where carriage returns usually do not work as you wish they would.

That is to say: this is quite nice for very small applications that can run on a personal computer, but does not translate to larger scale scenarios.

include/utils/conditional_o_stream_wrapper.hpp
```diff
@@ -0,0 +1,111 @@
...
+ * @brief A wrapper around dealii::ConditionalOStream
+ * that deletes warnings about the return value not being used.
+ *
+ * @details
+ * In CLion, using dealii::ConditionalOStream's operator<<
+ * generates warnings because the operator returns a reference
+ * to the stream, but this return value is not used.
+ * This wrapper class overrides the operator<< to
+ * return a reference to itself, effectively suppressing
+ * these warnings.
```
michelebucelli: This sounds like one more reason not to use `CLion` 😄 Does it complain for `cout` too? Since the syntax and semantics are the same. Jokes aside, this class seems like a complicated solution to what should be a simple problem.

include/time_integrator/newmark_integrator.hpp
```diff
@@ -0,0 +1,51 @@
...
+                 const AffineConstraints<>                       &constraints_u_np1,
+                 const std::map<types::global_dof_index, double> &u_boundary_values,
+                 const AffineConstraints<>                       &constraints_v_np1,
+                 const std::map<types::global_dof_index, double> &v_boundary_values,
+                 TrilinosWrappers::MPI::Vector                   &U,
+                 TrilinosWrappers::MPI::Vector                   &V) override;
+
+private:
+    double beta;
+    double gamma;
```
michelebucelli: These should definitely be const.

include/time_integrator/theta_integrator.hpp
```diff
@@ -0,0 +1,43 @@
...
+             const TrilinosWrappers::MPI::Vector             &F_np1,
+             const AffineConstraints<>                       &constraints_u_np1,
+             const std::map<types::global_dof_index, double> &u_boundary_values,
+             const AffineConstraints<>                       &constraints_v_np1,
+             const std::map<types::global_dof_index, double> &v_boundary_values,
+             TrilinosWrappers::MPI::Vector                   &U,
+             TrilinosWrappers::MPI::Vector                   &V) override;
+
+private:
+    double theta;
```
michelebucelli: This should be const.

src/time_integrator/theta_integrator.cpp
```diff
@@ -0,0 +1,134 @@
...
+    // Add: -K * U^n
+    TrilinosWrappers::MPI::Vector KU(U);
+    K.vmult(KU, U);
+    rhs.add(-1.0, KU);
+
+    const unsigned int max_it = 20000;      // << prima era 5000
+    const double       rel_tol = 1e-8;      // << per output/ParaView va benissimo
+
+    // Solve: C_LHS * V^{n+1} = rhs
+    SolverControl solver_control(max_it, rel_tol);
```
michelebucelli: This tolerance is not relative, despite what its name suggests! You should consider using the `ReductionControl` class instead.

include/functions/mms_functions.hpp
AndreVale69: Solved: https://github.com/AndreVale69/nmpde-wave-equation-solver/pull/21

include/time_integrator/newmark_integrator.hpp
AndreVale69: Solved: https://github.com/AndreVale69/nmpde-wave-equation-solver/pull/21

include/time_integrator/theta_integrator.hpp
AndreVale69: Solved: https://github.com/AndreVale69/nmpde-wave-equation-solver/pull/21

include/utils/conditional_o_stream_wrapper.hpp
AndreVale69: That's right. The wrapper has been removed. Solved: https://github.com/AndreVale69/nmpde-wave-equation-solver/pull/22

include/enum/boundary_type.hpp
AndreVale69: Thanks for the suggestion. We are currently targeting deal.II 9.5.1 (mk module), which does not yet bundle `magic_enum`. Since `magic_enum` is header-only, we can safely add it as an external dependency. We also agree that an exception should be thrown when an unknown enum value is encountered rather than a placeholder string being returned, and we have adopted that approach. Solved: https://github.com/AndreVale69/nmpde-wave-equation-solver/pull/24

include/parameters.hpp
AndreVale69: Solved: https://github.com/AndreVale69/nmpde-wave-equation-solver/pull/25

src/main.cpp
AndreVale69: Solved: https://github.com/AndreVale69/nmpde-wave-equation-solver/pull/26

include/utils/progress_bar.hpp
AndreVale69: Thanks for the interesting observation. We agreed that this approach was mainly suited for small-scale runs on personal machines and did not translate well to typical cluster workflows with schedulers and redirected output. We did not target or test the solver in a cluster context. For this reason, the progress bar was made fully optional and could be easily disabled, so that standard output remained compatible with batch and large-scale execution scenarios.

Solved: https://github.com/AndreVale69/nmpde-wave-equation-solver/pull/27

src/time_integrator/theta_integrator.cpp
AndreVale69: Solved: https://github.com/AndreVale69/nmpde-wave-equation-solver/pull/28

src/utils/mesh_generator.cpp
AndreVale69: Thanks professor for the suggestion and for pointing out the existing interfaces.

We investigated the available options in detail using deal.II v9.5.1 (the version provided by the MK module), and we would like to clarify why this is currently not planned.

**(1) deal.II and mesh generation from `.geo`**
In deal.II v9.5.1 we did not find a supported mechanism to *generate* `.msh` files starting from `.geo` inputs. The only related functionality we identified is the Gmsh utilities header
[https://github.com/dealii/dealii/blob/7e98d66134932ef70b1ae30929ece8488c049a55/include/deal.II/gmsh/utilities.h](https://github.com/dealii/dealii/blob/7e98d66134932ef70b1ae30929ece8488c049a55/include/deal.II/gmsh/utilities.h),
which provides helper utilities (e.g. handling physical tags) but does not perform geometry parsing or mesh generation. In practice, deal.II assumes that the `.msh` file is produced externally and then consumed via `GridIn`.

**(2) Use of the Gmsh C++ API**
We agree that, from a software engineering perspective, directly using the Gmsh C++ API would be a cleaner and more integrated solution, as it avoids invoking external processes and provides better control over geometry and meshing. However, we intentionally decided not to adopt this approach in the current project. Linking against `libgmsh` would introduce an additional SDK dependency that users would need to install and configure explicitly, with OS-specific binaries and build-system implications. Instead, the current approach only requires the Gmsh executable to be available in the environment, which is a lighter and more common requirement, especially in teaching, cluster, or evaluation environments.

For these reasons, we currently treat mesh generation as an external preprocessing step, while the solver itself only consumes `.msh` files. This choice is consistent with the standard deal.II workflow and favors portability and reproducibility over tighter library integration.

We agree that this is primarily a design and tooling decision rather than a functional limitation, and we will make sure to document this choice and its trade-offs clearly.

PR #46, opened 2025-12-23
NMPDE project: FSI Infascelli Rosnati Santoro

DarioSantoroDS:
Consegna del codice del progetto Fluid-Structure Interaction.


src/fsi.cpp
```diff
@@ -0,0 +1,84 @@
...
+#ifdef DEBUG
+      flow_problem.output_matrix(); // output matrix not anymore implemented, we used it to debug when PETSc was implemented and
+                                    // we could output the matrix to check it on matlab
+#endif
+#ifdef ITERATIVE_SOLVER
+      flow_problem.solve_iterative();
+#endif
+#ifdef DIRECT_SOLVER
+      flow_problem.solve(); // direct solver not anymore implemented, we used it when the matrix was not a block matrix
+#endif
```
michelebucelli, line 39: You should not use preprocessor directives for this kind of thing. This can be done as efficiently through both compile-time constants and runtime variables, with a much better syntax, robustness, support for errors in the compiler, etc.

src/FluidStructureProblem.hpp
```diff
@@ -0,0 +1,693 @@
...
+// If deal.II is in version lower than 9.4, the Assert
+// is a ExcNotImplemented, while in version 9.4 or higher it is an
+// AssertDimension that fails. Nevertheless, in Release mode, where the
+// Asserts are not active, the code works fine.
+// To avoid this problem, when in DEBUG mode
+// we do not use the block triangular preconditioner AMG, that needs the
+// constant modes to be extracted. Instead, we use a simpler block with AMG
+// preconditioner that does not use the constant modes.
+// It has to be said that that preconditioner is less efficient, but at least
+// it works.
```
michelebucelli, line 84: I have personally used the constant modes feature of the AMG preconditioner offered by Trilinos, so I am sure that, at least in a basic situation (no hp adaptivity, no hanging nodes) it works. Perhaps there are some conflicts with one of those features?

Either way, you should avoid workarounds such as this one in every way possible. The vast majority of the time assertions are there for a good reason.

src/FluidStructureProblem.hpp
```diff
@@ -0,0 +1,693 @@
...
+public:
+  ParameterReader(ParameterHandler &paramhandler)
+    : prm(paramhandler)
+  {}
+  void
+  read_parameters(const std::string &);
+
+private:
+  void
+                    declare_parameters();
```
michelebucelli, line 123: The indentation here is quite funny-looking and hard to read.

src/FluidStructureProblem.hpp
```diff
@@ -0,0 +1,693 @@
...
+#endif
+#ifdef ITERATIVE_SOLVER
+  void
+  solve_iterative();
+#endif
+  void
+  output_results(const unsigned int refinement_cycle) const;
+#ifdef DEBUG
+  void
+  output_matrix() const; // not anymore implemented
```
michelebucelli: If it is not implemented anymore, it should probably be removed. It is usually a good idea to keep your code in a clean state. If you need to recover it at a later time, you can always checkout an older git commit.

src/FluidStructureProblem.hpp
```diff
@@ -0,0 +1,693 @@
...
+
+    // Temporary vector stokes
+    mutable LA::MPI::Vector tmpStokes;
+  };
+
+  // Searching for other preconditioner to implement, we also tried a sort of
+  // SIMPLE preconditioner used for Navier-Stokes equations. However, the
+  // performance was not satisfactory, so we decided not to use it. The code is
+  // left here commented for possible future reference. class
+  // PreconditionBlockTriangularSimple
```
michelebucelli: This is interesting: why didn't you report the performance of this preconditioner in the report?

michelebucelli:
As a general comment, I notice that your code contains quite a bit of leftovers and commented-out parts. This is not a good practice, in my opinion: you should try to keep your code in a clean state, to avoid making mistakes due to accidental mixups with previous implementations, as well as to improve the readability, which is always a good thing.

In case you need to retrieve something from a previous iteration, you can always roll back to the commit that contained it (of course, this requires you being tidy in your commit history, which is also a good practice).

PR #47, opened 2025-12-29
Consegna progetto

lorenzozanizz:
Consegna del progetto NMPDE per proposta Project 6. Non-linear cardiac mechanics


src/cpp/main.cpp
```diff
@@ -0,0 +1,69 @@
...
+	Utilities::MPI::MPI_InitFinalize mpi_init(argc, argv);
+
+	// ENABLE ALL EXCEPTIONS FOR NAN GENERATION TO SEE WHERE THINGS GO ROGUE
+	feenableexcept(FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW);
+
+	std::vector<double> vals(100);
+	parse_params(argc, argv, vals);
+
+	// Show the name, a reference value and the real value taken as line argument
+#define p(s1, def, s) s
```
michelebucelli, line 42: Since we are in 2025 (well, almost 2026), you should try to avoid using preprocessor macros as much as possible - the C++ language is rich enough that you have safer and equally efficient alternatives.

In this particular case, if I get it right, the point of this macro is that each input argument to `OrthotropicSolver` is prefixed by its meaning and "default" value. This is quite obfuscated (at least by looking at lines 48-61) below: the use of the macro looks like a function call, so the reader may think that something's going on, although nothing is.

The appropriate way of doing that would have been through comments, e.g.
```cpp
OrthotropicSolver seis(
    /* Lagrange basis degree = */ (int) vals[0],
    /* p_v internal Neumann pressure = */ vals[1],
    /* ... */
  );
```

src/cpp/main.cpp
```diff
@@ -0,0 +1,69 @@
...
+	// example activation: 
+	//  mpirun -n 4 ./main 2 0.1 0.04 0.3 4.5 4.19 6.5 2.5 10.4 0.130 15.255 2.0 0.5 9.0
+
+	Utilities::MPI::MPI_InitFinalize mpi_init(argc, argv);
+
+	// ENABLE ALL EXCEPTIONS FOR NAN GENERATION TO SEE WHERE THINGS GO ROGUE
+	feenableexcept(FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW);
+
+	std::vector<double> vals(100);
+	parse_params(argc, argv, vals);
```
michelebucelli, line 39: You should either look into `dealii::ParameterHandler` and the associated classes, or a better parser for command line arguments (such as [clipp](https://github.com/muellan/clipp)), or both. This approach is extremely error prone - I imagine it's not too hard for a user to swap two command line arguments by mistake.

src/cpp/solver/boundaries.hpp
```diff
@@ -0,0 +1,20 @@
...
+#ifndef __SOLVER_BOUNDARIES
+
+/* =========== PREVIOUS MESH VALUES ===========
+#define PDE_ROBIN		2
+#define PDE_DIRICHLET	3
+#define PDE_NEUMANN		4
+*/
+#define PDE_ROBIN		1
+#define PDE_DIRICHLET	2
+#define PDE_NEUMANN		3
```
michelebucelli: I gather from what follows that these are meant to be the boundary IDs for the different boundaries. In which case,
1. the macro name is not very expressive (PDE stands for partial differential equation: why would these be boundary IDs?)
2. you should not have used a macro, but rather a member of the solver class (maybe `const`, maybe `constexpr`, maybe neither, but in any case a C++ variable), which would give you type safety and a number of other nice features that you don't have with preprocessor macros. Really, don't use preprocessor macros unless there really is no other way, and most of the time there is.

src/cpp/utilities/refine.hpp
```diff
@@ -0,0 +1,47 @@
...
+namespace UtilsMesh {
+
+    using namespace dealii;
+
+
+    /* @brief metafunction for refinement of a mesh up to n steps of refinement.
+
+     @param triangulation the mesh into a triangulation
+     @param steps the amount of refinement steps
+     @param save_into where the mesh is to be saved as a .vtu file
```
michelebucelli, line 20: This will not be a VTU file, but an MSH file. "No documentation is better than wrong documentation"

src/cpp/utilities/refine.hpp
```diff
@@ -0,0 +1,47 @@
...
+     @tparam D the dimension of the mesh
+    */
+    template <int D>
+    void
+    refine_dmesh_nsteps(
+        Triangulation<D>& triangulation,
+        const unsigned int steps,
+        const std::string& save_into
```
michelebucelli, line 28: This is admittedly a minor issue, but this function has a bit of a design problem.

The name and the documentation of the function say that this does precisely one thing: it applies n steps of refinement to a mesh. However, the body of the function does two things: it applies the refinement and it writes the mesh to file (this is hinted at by the input arguments, but not really documented). Thus, you have a function whose actual behavior contrasts with the expectations that are set by its name, its documentation, and its overall outward appearance. 

This is bad practice: if you do this kind of thing in a larger code base, things may get out of hand pretty quickly.

In this particular case, say that someone in the future (maybe yourselves) uses this function, and finds that it is super slow. They might think that there's something wrong with their mesh, or with its refinement, while the issue may simply be that writing to disk takes a bit of time, and the issue lies in IO. This may take a while for them to find out.

General-purpose functions such as this one should always be designed very carefully.

src/cpp/vectorize/matrix_op.cpp
```diff
@@ -0,0 +1,43 @@
+#include "immintrin.h"
+
+FORCE_INLINE
+void 
+do_voigt_product(
```
michelebucelli, line 5: If I understand correctly, here you're implementing vectorized operations. Have you, by any chance, looked into the `dealii::VectorizedArray` class?

michelebucelli:
As a general comment, the code is quite hard to read and a bit convoluted.

Some of this is related to the use of old idioms (such as preprocessor directives) which should be replaced by more modern C++ language features.

Other times, the variable and function names that you use are not very expressive, which makes it hard to understand their purpose. I'm thinking for instance of `pde_out_c_par`, `pass_cache_data_t`, and a few other cases. It is generally a good practice to use self documenting names wherever possible.

In some other places, the behavior of the program appears a bit obfuscated, e.g. in the use of the function `compute_and_cache` during assembly. This may be motivated by optimization reasons (although I did not see any proof that this is useful or beneficial in the report, so I cannot really comment on that), but it should probably be wrapped inside some more expressive syntax.

PR #49, opened 2025-12-29
Project 6: Code delivery

kuchpio:
# Project 6: Non-Linear Cardiac Mechanics
Group: Kucharczyk, Matar, Ostyk-Narbutt, Lenkšas


src/neo-hookean-cube/src/NeoHookeCube.hpp
```diff
@@ -0,0 +1,146 @@
+#ifndef NON_LINEAR_DIFFUSION_HPP
+#define NON_LINEAR_DIFFUSION_HPP
```
michelebucelli: This headerguard appears to be outdated.

michelebucelli:
My main criticism of this is that you have a lot of duplicated code between the different test cases. 

The cube, beam and heart cases differ "only" in their mesh, boundary conditions and constitutive law (that is, the function that relates the stress tensor to the displacement). Significant parts of the respective solvers overlap. Not only code duplication is generally a bad thing, but also all your test case are logically particular instances of a more general "concept" (a non-linear mechanics solver).

Accordingly, the code would have been better structured with a "mechanics" module that manages the common parts, and in which the different meshes, boundary conditions and constitutive models can be plugged in. The different test cases would just be particular instances of this module.

In practice, there's a few different ways in which this could be achieved. For instance, the constitutive law could be passed in as a lambda function (or any function object) or as a polymorphic class. The same goes for the mesh generation.

